PR #130, opened 2022-08-29
CQ-4345916 Capture DOR specific properties for the core components

devgurjar:
@Review : @vdua @nitigupt
Added code to support dor properties on element level

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Field.java
+++ b/bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Field.java
@@ -18,6 +18,8 @@ package com.adobe.cq.forms.core.components.models.form;
 import org.osgi.annotation.versioning.ConsumerType;
 
 import com.adobe.cq.forms.core.components.util.DefaultValueSerializer;
+import com.adobe.cq.forms.core.components.views.Views;
+import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
 /**
@@ -91,6 +93,16 @@ public interface Field extends Base, BaseConstraint {
     @JsonSerialize(using = DefaultValueSerializer.class)
     default Object[] getDefault() {
         return null;
+    }
+
+    @JsonView(Views.Author.class)
+    default boolean getDorExclusion() {
+        return false;
+    }
+
+    @JsonView(Views.Author.class)
+    default String getDorColspan() {
+        return "";
     }
 
 }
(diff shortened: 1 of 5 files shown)


bundles/core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFieldImpl.java
```diff
@@ -104,6 +104,14 @@ public abstract class AbstractFieldImpl extends AbstractBaseImpl implements Fiel
     @SlingObject
     private Resource resource;
 
+    @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL)
+    @javax.annotation.Nullable
+    protected boolean dorExclusion;
+
+    @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL)
```
nit23uec, line 111: as discussed today lets keep DoR specific additional props in custom properties.

bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Field.java
```diff
@@ -93,4 +95,14 @@ default Object[] getDefault() {
         return null;
     }
 
+    @JsonView(Views.Author.class)
+    default boolean getDorExclusion() {
```
rismehta, line 99: This needs to be part of fd:dor namespace as custom properties. You can check existing implementation of custom properties related to layout, https://github.com/adobe/aem-core-forms-components/blob/dev/bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Base.java#L340-L349

In a similar manner we could add getDorProperties in base class with JsonView set to Author.

rismehta: This would have jsonIgnore annotation, since it would be part of `fd:dor` namespace. Each piece of code in core components needs to have proper java doc.

rismehta:
Please check comments

rismehta (changes requested):

rismehta:
PR needs to be raised from same branch

bundles/core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFieldImpl.java
devgurjar: Sure

bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Field.java
devgurjar: Sure, I will incorporate the comments.

PR #170, opened 2022-09-14
CQ-4345916 Capture DOR specific properties for the core components

devgurjar:
@Review : @rismehta @vdua @nitigupt
@dod : done
Added support of dor and added test cases

## Types of changes

## Checklist:

### DIFF
--- a/bundles/core/src/test/resources/form/dropdown/exporter-multiselect-dropdown.json
+++ b/bundles/core/src/test/resources/form/dropdown/exporter-multiselect-dropdown.json
@@ -11,7 +11,10 @@
   "properties": {
     "af:layout" : {
       "tooltipVisible": false
-    }
+    },
+    "fd:dor":{
+      "dorExclusion":false,
+      "dorColspan":null}
   },
   "enforceEnum":false,
   "enum" : [0, 1, 2],
--- a/bundles/core/src/test/resources/form/panel/exporter-rules-panel.json
+++ b/bundles/core/src/test/resources/form/panel/exporter-rules-panel.json
@@ -1,46 +1,58 @@
 {
-  "id": "panel-81f441fcab",
+  "id":"panel-81f441fcab",
+  "description":"dependent names description",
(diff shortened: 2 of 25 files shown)


bundles/core/src/test/resources/form/dropdown/exporter-multiselect-dropdown.json
```diff
@@ -11,7 +11,10 @@
   "properties": {
     "af:layout" : {
       "tooltipVisible": false
-    }
+    },
+    "fd:dor":{
+      "dorExclusion":false,
+      "dorColspan":null}
```
vdua, line 17: `dorcolspan` is not a number ?

vdua (approved):

bundles/core/src/test/resources/form/dropdown/exporter-multiselect-dropdown.json
devgurjar: It is number but input type is text.

bundles/core/src/test/resources/form/panel/exporter-rules-panel.json
```diff
@@ -1,46 +1,58 @@
...
+        "visible":true,
+        "richText":false,
+        "value":"def"
       },
-      "type": "string",
-      ":type": "core/fd/components/form/textinput/v1/textinput"
+      "properties":{
+        "fd:dor":{
+          "dorExclusion":false,
+          "dorColspan":null
```
rismehta, line 29: Why are we sending this when it is null. We should only include in json if it is not null

devgurjar: By default in v1, dorExclusion and dorColspan contains no value. It gets value only if we explicitly set from authoring. So I guess this shouldn't be part of json by default.

rismehta: Irrespective of author/publish view, non-nullable values should only be part of json.

PR #173, opened 2022-09-15
CQ-4347957 Prefill Support

nit23uec:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/HTTPAPILayer.js
+++ b/ui.frontend/src/HTTPAPILayer.js
@@ -13,19 +13,42 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  ******************************************************************************/
+import {Constants} from "./constants";
 
 export default class HTTPAPILayer {
 
+    static async getForm(formContainerPath) {
+        const _formPath = formContainerPath.substring(0, formContainerPath.indexOf("/jcr:content/guideContainer"));
+        const _formsList =  await this.#getFormsList();
+        if (_formsList) {
+            return await this.#findForm(_formPath, _formsList);
+        } else {
+            throw new Error("Error in fetching form");
+        }
+    }
+
     static async getFormDefinition(formId) {
-        return await this.#getJson("/adobe/forms/af/v1/" + formId);
+        return await this.#getJson(Constants.API_PATH_PREFIX + "/" + formId);
+    }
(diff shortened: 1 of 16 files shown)


ui.frontend/src/utils.js
```diff
@@ -140,10 +124,22 @@ export default class Utils {
             if (_path == null) {
                 console.error(`data-${Constants.NS}-${formContainerClass}-path attribute is not present in the HTML element. Form cannot be initialized` )
             } else {
-                const _formJson = await Utils.getJson(_path + ".model.json");
-                console.log("model json from server ", _formJson);
+                const _formsList = await HTTPAPILayer.getFormsList();
```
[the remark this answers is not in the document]
nit23uec: we should not expose the id generation logic in public.

ui.frontend/src/view/FormContainer.js
```diff
@@ -20,6 +20,9 @@ export default class FormContainer {
         this._path = params._path;
         this._fields = {};
         this._deferredParents = {};
+        if (params._prefillData) {
+            this._model.importData(params._prefillData.data);
+        }
```
vdua, line 25: I think better would be 
`createFormInstance({...params._formJson, data: _params._prefillData})`

vdua:
Since it requires latest aemds (af2-rest-api), moving it to draft state.

ui.frontend/src/HTTPAPILayer.js
```diff
@@ -0,0 +1,72 @@
...
+        } else {
+            //TODO: throw errors once API is available on Circle CI set up
+            console.debug("Error in fetching form");
+        }
+    }
+
+    static async getFormDefinition(formContainerPath) {
+        return await this.#getJson(formContainerPath + ".model.json");
+    }
+
```
vdua, line 34: I am still not comfortable with making so many network calls to get the form definition.

nit23uec: there's still only one call required to fetch the model definition.
but for prefill call, we need form id, and hence, we will need listForms call.
issue is already logged to optimise it. 
https://github.com/adobe/aem-core-forms-components/issues/190

vdua (approved):

PR #175, opened 2022-09-15
CQ-4346666 Ensure schema compliance of sling models as per AF 2.0 spe…

rismehta:
…cification

@review @vdua
DOD(yes)

## Types of changes

## Checklist:


bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -64,6 +64,11 @@ public String getThankYouMessage() {
         return thankyouMessage;
     }
 
+    @Override
+    public String getAdaptiveFormVersion() {
```
nit23uec, line 68: this should be picked up from storage (jcr here).

bundles/core/src/test/resources/schema/adaptive-form-properties.schema.json
```diff
@@ -0,0 +1,322 @@
+{
+  "$id": "classpath:/schema/adaptive-form-properties.schema.json",
```
nit23uec, line 2: this approach is not scalable. I'm assuming that we will fetch the publicly published schema in future.

nit23uec (changes requested):
I see this flow like:
1. AF gets created with a particular schema version (which gets persisted on guide container).
this schema version is the same for which runtime (af-core) is there in AEM build.
2. Now the json generated from the sling models on that AEM build MUST comply to that particular schema version.
3. I can imagine a need of supporting multiple schema versions in the code (given the fact that we will have to support AFs created from all schema versions)

Do we intent to do the above task as part this jira or is there a separate action item for the same?

bundles/core/src/test/resources/schema/adaptive-form-properties.schema.json
rismehta: Yes, this is true for even the VS code plugin. Till the schema is public, we maintain a copy.

bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
rismehta: No. Any changes in spec would mean a new version of form container. Hence the value needs to be hard coded

rismehta:
> Now the json generated from the sling models on that AEM build MUST comply to that particular schema version

Right now, we don't validate the json at runtime(since we don't have to support to validate v7 json schema in  AEM yet

> I can imagine a need of supporting multiple schema versions in the code (given the fact that we will have to support AFs created from all schema versions)

I have changed the test case to incorporate schema version.

bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
nit23uec: spec follows semantic versioning. the changes in minor/patch versions should not need another version of core component.

rismehta: That depends on the use-case. if we need to a new component, we would add it, else we could add the changes in the same component.

barshat7 (approved):

PR #192, opened 2022-09-22
Base Class for Static Components

vdua:
## Description

Creating a base class for static components which can be extended by Text and Image Component

### DIFF
--- a/bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Base.java
+++ b/bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Base.java
@@ -18,22 +18,13 @@ package com.adobe.cq.forms.core.components.models.form;
 import java.util.Collections;
 import java.util.Map;
 
-import javax.annotation.Nonnull;
-
-import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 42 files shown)


bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/form/Base.java
```diff
@@ -18,221 +18,23 @@
...
     public final String DATE_FORMATTER = "yyyy-MM-dd";
 
-    /**
-     * Defines the view type. Possible values: {@code text-input}, {@code multiline-input}, {@code number-input}, {@code date-input},
-     * {@code file-input}, {@code drop-down}, {@code radio-group}
-     * , {@code plain-text}, {@code checkbox}, {@code button}, {@code panel}
-     *
-     * @since com.adobe.cq.forms.core.components.models.form 0.0.1
-     */
-    public enum FieldType {
```
vdua, line 56: Moving this to Base to FormComponent would require a major version bump (as it changes the bytecode in a backwards incompatible way). Hence we are moving it to a separate place only

nit23uec (approved):

rismehta (approved):

PR #275, opened 2022-10-18
Supporting CSRF while making a request

vdua:
## Types of changes

## Checklist:


ui.frontend/clientlib.config.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
       ...libsBaseConfig,
       name: 'core-forms-components-runtime-base',
       categories: ['core.forms.components.runtime.base'],
+      dependencies: ['core.forms.components.csrf'],
```
rismehta, line 49: You can use the `granite.csrf.standalone` client lib for this use-case

ui.frontend/src/csrfRequest.js
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+import {FunctionRuntime, request} from '@aemforms/af-core'
+
+export const enableCSRF = () => {
+    FunctionRuntime.registerFunctions(window.Granite.csrf.request(request))
```
nit23uec, line 20: we need to update af-core library first before merging this. else all runtime test cases will fail.

ui.frontend/src/customFunctions.js
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ ******************************************************************************/
+
+/**
+ *
+ * @param str {string} json string to convert custom function to object
+ * @return {object} JSON Object after parsing the string as json. In case of
+ * exceptions empty object is returned
+ */
+function toObject(str) {
```
nit23uec, line 23: these are defined at global level right now. we should put them under scope.

ui.frontend/src/customFunctions.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+function externalize(url) {
+    if (window?.Granite?.HTTP && typeof (window.Granite.HTTP.externalize === "function")) {
+        return window.Granite.HTTP.externalize(url);
+    } else {
+        return url;
+    }
+}
+
+export const customFunctions = {
```
nit23uec, line 45: are these  custom functions really needed here? can we defined them in json-formula instead?

nit23uec (changes requested):
custom functions can be avoided if we are overriding native fetch.

ui.frontend/src/customFunctions.js
vdua: externalize would need to be defined somewhere outside of json formula as it is aem specific. I also wanted these to be examples so that more functions can be registered.

Raised an issue in JSON Formula for toObject (or similar)

https://github.com/adobe/json-formula/issues/35

ui.frontend/src/customFunctions.js
vdua: That is a limitation of our custom functions right now.

vdua: But since they are not exported, they would not pollute the window namespace.

nit23uec (approved):

PR #291, opened 2022-10-28
CQ-4348632 Fixed Boolean enum type not working in Dropdown

ravise5:
## Types of changes

## Checklist:


ui.apps/src/main/content/jcr_root/apps/core/fd/components/form/dropdown/v1/dropdown/dropdown.html
```diff
@@ -42,10 +42,9 @@
             data-sly-attribute.multiple="${dropdown.isMultiSelect ? 'multiple' : ''}">
         <option data-sly-test="${dropdown.placeHolder != null}" value="" disabled selected>${dropdown.placeHolder}</option>
         <div data-sly-list.enumName="${dropdown.enumNames}" data-sly-unwrap>
-            <div data-sly-test.enum="${dropdown.enums[enumNameList.index]}" data-sly-unwrap>
-                <option value="${enum}"
+                <option value="${dropdown.enums[enumNameList.index].toString}"
```
nit23uec, line 45: can `dropdown.enums[enumNameList.index]` be null?

nit23uec: I think we should keep the `data-sly-test ` to account for null (just not assign it to any variable)

ravise5: `dropdown.enums[enumNameList.index]`  is null only when we allow the user to skip either the save value or display value in the edit dialog. 
 Checks for which ig needs to be in place. 
 Currently radiobutton and checkbox also uses this same approach

ravise5: Maybe in a prefill use-case ig it is possible to have null values, but then again we are not doing this check currently for enumNames.

nit23uec (approved):

PR #309, opened 2022-11-02
CQ-4348882 showing default error message in case some condition fails

im-shiv:
No default error message when required message is not added

### DIFF
--- a/ui.frontend/src/view/FormFieldBase.js
+++ b/ui.frontend/src/view/FormFieldBase.js
@@ -179,7 +179,7 @@ export default class FormFieldBase extends FormField {
 
     _updateErrorMessage(errorMessage, state) {
         if (this.errorDiv) {
-            this.errorDiv.innerHTML = state.errorMessage;
+            this.errorDiv.innerHTML = state.errorMessage || 'There is an error in the field';
         }
     }
 


nit23uec:
radiobutton and checkbox grp TCs are failing due this change. @im-shiv can you ps have a look?

ui.tests/test-module/specs/radiobutton/radiobutton.runtime.spec.js
```diff
@@ -99,7 +99,7 @@ describe("Form with Radio Button Input", () => {
         cy.get(`#${id}`).find(".cmp-adaptiveform-radiobutton__errormessage").should('have.text',"This is a required radiobutton");
 
         cy.get(`#${id}`).find("input").eq(1).click().then(x => {
-            cy.get(`#${id}`).find(".cmp-adaptiveform-radiobutton__errormessage").should('have.text',"");
+            cy.get(`#${id}`).find(".cmp-adaptiveform-radiobutton__errormessage").should('have.text',"There is an error in the field");
```
nit23uec, line 102: this seems incorrect. we should fix `_updateValid` method also  in case error msg is defined in the edit dialog to override the default msg.

im-shiv: that is the case

nit23uec (approved):

PR #325, opened 2022-11-10
CQ-4349511 Fixed name field in dialog not checking for invalid characters

ravise5:
## Types of changes

## Checklist:

### DIFF
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/base/v1/base/clientlibs/editor/js/editDialog.js
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/base/v1/base/clientlibs/editor/js/editDialog.js
@@ -122,6 +122,22 @@
         }
     }
 
+    function validateName() {
+        var elementNameRegex = /^[\w\-]+$/;  // word char [A-Za-z0-9_] and '-'
+        // Validator for name field. Please use node-name-validation where we are creating a node with name using this field
+        $(window).adaptTo("foundation-registry").register("foundation.validation.validator", {
+            selector : "[data-node-name-validation]",
+            validate : function (el) {
+                var name = el.value;
+
+                if (!elementNameRegex.test(name)) {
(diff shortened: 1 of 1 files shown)


ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/base/v1/base/clientlibs/editor/js/editDialog.js
```diff
@@ -122,6 +122,22 @@
         }
     }
 
+    function validateName() {
```
nit23uec, line 125: We are not creating any jcr bode with the name option. The validation should align with the AF2 spec rather than the jcr name validation

ravise5: https://jira.corp.adobe.com/browse/CQ-4349511
As per @vdua in the above ticket as per AF2 spec only restriction on name field is that it shouldn't begin with `$`.
@nit23uec I hope this is what you are trying to say here..

vdua: The problem with that is the name will not work in DoR as per the AF Spec currently. We need to fix that in DoR before removing this restriction. We have to make that work before removing this restriction.

vdua (approved):

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/base/v1/base/clientlibs/editor/js/editDialog.js
vdua: Also the rule editor would need some changes as well while transforming special names.

ravise5:
As discussed in the scrum logged a ticket for aligning element name with AF2 spec
https://jira.corp.adobe.com/browse/CQ-4349643
cc: @nit23uec

PR #327, opened 2022-11-14
CQ-4348721 Fixed number input for safari browsers

ravise5:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/utils.js
+++ b/ui.frontend/src/utils.js
@@ -206,4 +206,11 @@ export default class Utils {
         }
         return prefillJson;
     }
+
+    /**
+     * detects safari browser
+     */
+    static isSafari() {
+        return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+    }
 }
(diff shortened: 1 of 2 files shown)


ui.frontend/src/utils.js
```diff
@@ -206,4 +206,11 @@ export default class Utils {
         }
         return prefillJson;
     }
+
+    /**
+     * detects safari browser
+     */
+    static isSafari() {
+        return /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
```
nit23uec, line 214: we are trying to make ui.frontend  module to be as isomorphic as possible.
pls don't assume that window.navigator would be available in the client.
additional null check for navigator is required.
method name can be `isUserAgent(agentName)`

nit23uec (changes requested):
inline comment

ui.frontend/src/utils.js
ravise5: Done.

nit23uec (approved):

PR #332, opened 2022-11-17
[CQ-4348956_2] added changes to read from models getters rather than …

deepprakash345:
added changes to read from models getters rather than from getState api which returns the _jsonModel. getters have been modified to honor their parents enable property as well.

Testing.

- Verified with individual component directly dropped in form.
- Verified with individual component dropped in wizard. All 4  permutations of wizard being disabled/enabled and the components being disabled/enabled was tested and verified.
- Verified with a panel inside a wizard . And then dropping components inside parent. This verifies the layout component inside layout component scenario.

corresponding pr of af2-web= https://git.corp.adobe.com/livecycle/af2-web-runtime/pull/367
Jira:- https://jira.corp.adobe.com/browse/CQ-4348956

@vdua , @nit23uec Please review.


ui.frontend/src/view/FormFieldBase.js
```diff
@@ -99,7 +99,7 @@ export default class FormFieldBase extends FormField {
             this._updateValue(state.value);
         }
         this._updateVisible(state.visible)
-        this._updateEnabled(state.enabled)
+        this._updateEnabled(this._model.enabled)
```
vdua, line 102: I am not sure why are we doing this. Reading this way, enables dependency tracking which we do not want.

In the core you need to change the enabled property in the state as well. getState() already does that for some properties which are dynamic.

vdua: We always update the _jsonModel and I understand we do not want that for the enabled property but the getter and the getState() should always return the same value.

deepprakash345: Ok @vdua , thats what i wanted to understand that whether _jsonModel is meant to always store the values as saved in crx or its values can change during runtime accordingly. 

I will make the change, since we are not actually altering the enabled field in model and only the getter is altering the returned value, should i change the value of _jsonModel.enabled only when getState() is called?
cc:- @vdua , @nit23uec

deepprakash345: Also @vdua , 
just for my understanding, can you please explain why we shouldn't use this._model.enable, how is it dependency tracking in this way. i am clear that getState and getters should return the same thing, that's what i also discussed yesterday but i guess i understood the solution incorrectly.
Thanks
Deep

PR #343, opened 2022-11-22
CQ-4349547 Sites Accordion does not work with AF 2.0.

sakshi-arora1:
## Types of changes

## Checklist:


it/apps/src/main/content/jcr_root/apps/forms-core-components-it/clientlibs/clientlib-it-base/.content.xml
```diff
@@ -2,4 +2,5 @@
 <jcr:root xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:cq="http://www.day.com/jcr/cq/1.0"
     jcr:primaryType="cq:ClientLibraryFolder"
     allowProxy="{Boolean}true"
-    categories="[forms-it.base]"/>
+    categories="[forms-it.base]"
```
nit23uec, line 5: how are we fixing it in the product?
I think we should include them in the page component.

sakshi-arora1: We don't ship any template in product, and these clientlibs should be referred via template policy. (same is with sites). We should not include anything in page component apart from ui.frontent module (that encapsulates af2-web-runtime). Rest all the core component libraries should be include via template page policies. This needs correction as well.

PR #357, opened 2022-11-30
Testing parallelisation and test splitting in circleci 

ravise5:
## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,9 @@ common:
           at: /home/circleci/build
       - run:
           name: UI tests
-          command: node .circleci/ci/it-tests.js
+          command: |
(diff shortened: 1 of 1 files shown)


ravise5:
This is a test PR. 
One concern is we are passing the globe of test files to the `it-test.js` script inside which we are using a shell script for starting the `mvn` command. So I'm not sure if parallelisation will occur. 
So for time being I've limited the number of executors to be spun up to only 4.

ravise5:
I've solved the problem of all tests getting executed on all spun up executors by passing the split test info as command line arguments to the script `it-test.js` and then modifying the pom.xml of ui.tests to accept these as an arguments which can be then passed to the  `cypress run --spec=""` command. 
But for some reason this command is erroring out inside the ci but working fine locally. Currently debugging this part.

PR #360, opened 2022-12-01
PR related to movement of formatters to design dialog and converting formatters to unicode encoded format.

deepprakash345:
WIP PR. Do not merge. Just want to get an early review on the approach.

Still need to make changes in text and datepicker.

Need to write test cases.

Just want to get the approach reviewed early.

cc:- @rismehta , @nit23uec , @vdua


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
```diff
@@ -131,57 +135,49 @@ protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHtt
...
             actionTypeDataSource = new SimpleDataSource(getDataSourceResources(
-                request.getResourceResolver(), type, dataModel, fieldType).iterator());
+                request, type, dataModel).iterator());
         }
         request.setAttribute(DataSource.class.getName(), actionTypeDataSource);
     }
 
-    private List<Resource> getDataSourceResources(ResourceResolver resourceResolver, FormMetaDataType type, String dataModel,
-        String fieldType) {
+    private List<Resource> getDataSourceResources(SlingHttpServletRequest request, FormMetaDataType type, String dataModel) {
```
rismehta, line 144: Can we use HttpServletRequest here ? Also, let's pass the resource resolver explicitly as argument.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
```diff
@@ -131,57 +135,49 @@ protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHtt
...
-                        if (formatterResource != null) {
-                            String path = formatterResource.getPath();
-                            ValueMap formatterResourceValueMap = formatterResource.getValueMap();
-                            Iterator<Map.Entry<String, Object>> it = formatterResourceValueMap.entrySet().iterator();
-                            while (it.hasNext()) {
-                                Map.Entry<String, Object> entry = it.next();
-                                String key = entry.getKey();
-                                // This condition is sufficient for property guideComponentType to get ignored
-                                if (key.startsWith("pattern")) {
+                    Resource contentResource = resourceResolver.getResource((String) request.getAttribute(Value.CONTENTPATH_ATTRIBUTE));
```
rismehta, line 152: Can we create a utility to fetch policies so that code could be re-used ?

nit23uec:
overall approach lgtm!

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
deepprakash345: done

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
deepprakash345: done

deepprakash345:
Verified all the formatters of datepicker in playground.
Formatters worked when the date-picker was converted to textfield. So in core-components we will have to add the corresponding date-picker changes for formatters to work in runtime.
NumberInput formatters are not yet verified as the same was not implemented at the react components side for now.

cc:- @nit23uec , @vdua

nit23uec (approved):
merging this as edit dialogs are broken on latest set ups.

PR #386, opened 2023-01-09
CQ-4349886 GuideBridge API to hide buttons

nit23uec:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/GuideBridge.js
+++ b/ui.frontend/src/GuideBridge.js
@@ -197,15 +197,49 @@ export default class GuideBridge {
         }
     }
 
+    /**
+     * @summary Hides all the submit buttons present in the Adaptive Form
+     *
+     */
     hideSubmitButtons() {
-        //TODO: implement it later. NO-OP for now.
+        if (this.isConnected()) {
+            this.#hideButtons('submit');
+        } else {
+            throw new Error("GuideBridge is not connected");
+        }
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


ui.frontend/src/GuideBridge.js
```diff
@@ -197,15 +197,49 @@ export default class GuideBridge {
...
+        }
     }
 
-    hideResetButtons() {
+    #hideButtons(buttonType) {
+        let formModel = this.getFormModel();
+        formModel.visit((field) => {
+            if (field.properties["fd:buttonType"] === buttonType) {
+                field.visible = false;
+                field.subscribe((action) => {
```
rismehta, line 229: Why do we need subscribe here ?

nit23uec: we want to hide the submit button always if this API is called.
even in the case someone has written a conditional rule to show it.

rismehta (approved):

ui.frontend/src/GuideBridge.js
deepprakash345: we are not checking for the buttonType to be "submit" . do we need to add the subscribe code in case buttonType is submit or for all buttonTypes?

nit23uec: only buttontype field is subsribed.

nit23uec:
2 flaky tests failed.

PR #389, opened 2023-01-09
Cq 4348947

barshat7:
## Description

If Screen Reader Preference is set by the author, then corresponding aria-label will be set for the component.

## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractOptionsFieldImpl.java
```diff
@@ -46,6 +46,10 @@ public abstract class AbstractOptionsFieldImpl extends AbstractFieldImpl impleme
     @Nullable
     private String[] enumNames;
 
+    @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL, name = "custom")
```
nit23uec, line 49: why name is set as custom in annotations?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
```diff
@@ -177,16 +177,20 @@ public String getScreenReaderText() {
     @Nullable
     public String getHtmlScreenReaderText() {
         // this can be used in sightly to compute initial html
-        String screenReaderText = getName();
+        String screenReaderText = "";
         if (AssistPriority.LABEL.equals(assistPriority)) {
```
nit23uec, line 181: is there a default assist priority value desired now? It does not seems so from this PR.
we should have a default value for screen reader text.

nit23uec:
why do we need `aria-label` in all core components?
it's required only when there is no appropriate text visible in the DOM that could be referenced as a label. Do we have any such core component?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
barshat7: Default is empty. (No corresponding property is saved in JCR)

barshat7:
> why do we need `aria-label` in all core components? it's required only when there is no appropriate text visible in the DOM that could be referenced as a label. Do we have any such core component?

Screenreader preference could be different than the <label> e.g Description or Custom Text, as we won't be able to change the label of the field, we are resorting to ARIA attributes. I think this is on par with V1. @rismehta

PR #436, opened 2023-01-27
Add themeclient lib 65

suryansh29:
## Types of changes

## Checklist:

- [x] My change requires a change to the documentation.


ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/page/v1/page/customheaderlibs.html
```diff
@@ -18,6 +18,11 @@
      data-sly-test.clientLibRef="${formstructparser.clientLibRefFromFormContainer}">
     <sly data-sly-test="${clientLibRef}" data-sly-call="${clientlib.css @ categories=clientLibRef}"/>
 </sly>
+<sly data-sly-use.clientlib="core/wcm/components/commons/v1/templates/clientlib.html"
+     data-sly-use.formstructparser="com.adobe.cq.forms.core.components.models.form.FormStructureParser"
```
nit23uec, line 22: how would it work for multiple form container in a page?

nit23uec (changes requested):
Approach wouldn't work for multiple form containers in a page.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/page/v1/page/customheaderlibs.html
suryansh29: Do we even need to have different themes in multiple form containers inside a page?
Also, even in master we are applying theme via page component itself through slingcontext aware config

Also, there is no option in site editor to have a different theme for form other than site.

suryansh29: cc @pankaj-parashar

suryansh29: cc @sakshi-arora1 @pankaj-parashar

pankaj-parashar:
> Approach wouldn't work for multiple form containers in a page.

How this is working for front-end theme approach? as the same case would be there, and theme config is being read at https://github.com/adobe/aem-core-wcm-components/blob/main/content/src/content/jcr_root/apps/core/wcm/components/page/v3/page/head.html#L45
In a site page if customer using multiple form containers, a single theme will be used ideally and that client library should be used in page component. In AF page, do we support multiple containers? If yes, how we are dealing this in front-end theme, as only one theme can be applied to a form.

nit23uec:
> > Approach wouldn't work for multiple form containers in a page.
> 
> How this is working for front-end theme approach? as the same case would be there, and theme config is being read at https://github.com/adobe/aem-core-wcm-components/blob/main/content/src/content/jcr_root/apps/core/wcm/components/page/v3/page/head.html#L45 In a site page if customer using multiple form containers, a single theme will be used ideally and that client library should be used in page component. In AF page, do we support multiple containers? If yes, how we are dealing this in front-end theme, as only one theme can be applied to a form.


As this file (page RT) is a common file in case of both sites and forms editor (as template may use this component), so we would want to make sure that this is a NO-OP with no side effects.
forms themes should be fetched as per sites clientlibs in any approach for forms in sites.

suryansh29:
@nit23uec Tested this with Form inside a site page. This code block is actually not triggered in that scenario, it will have no impact there.

nit23uec (approved):

PR #441, opened 2023-01-30
Cq 4350599, load clientlibs in iframe mode in aem forms container

deepprakash345:
Added clientlibs of core components in AEMForm Container in iframe mode.
Exposed isAdaptiveFormV2 api in AEMFormImpl.
Exposed registerConfig method in guidebridge.

Added testcases.


bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/aemform/AEMFormImpl.java
```diff
@@ -278,4 +285,22 @@ public String getHeight() {
         }
         return height;
     }
+
+    private String getFormVersion(ResourceResolver resourceResolver, String formPath) {
```
nit23uec, line 289: why do need resourceresolver and formpath in args here? it should already be present in the model

bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/aemform/AEMFormImpl.java
```diff
@@ -262,6 +264,11 @@ public boolean isFormSelected() {
         return !FormType.NO_FORM_SELECTED.equals(getFormType());
     }
 
+    @Override
+    public boolean isAdaptiveFormV2() {
```
nit23uec, line 268: I'd prefer `getVersion` API exposed instead of exposing APIs for every different version in future.

bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/aemform/AEMForm.java
```diff
@@ -40,7 +40,7 @@ public interface AEMForm extends Component {
      * Defines the form type
      */
     enum FormType {
-        NO_FORM_SELECTED, MOBILE_FORM, ADAPTIVE_FORM, MC_DOCUMENT, MOBILE_FORMSET
+        NO_FORM_SELECTED, MOBILE_FORM, ADAPTIVE_FORM, MC_DOCUMENT, MOBILE_FORMSET, ADAPTIVE_FORM_V2
```
nit23uec, line 43: FormType should be ADAPTIVE_FORM only. 
version would be the differentiator.

nit23uec (changes requested):
inline comments.

ui.frontend/src/GuideBridge.js
```diff
@@ -247,6 +249,10 @@ export default class GuideBridge {
         //TODO: implement it later. NO-OP for now.
     }
 
+    registerConfig(key, config) {
```
nit23uec, line 252: js docs pls.

ui.frontend/src/GuideBridge.js
```diff
@@ -247,6 +249,10 @@ export default class GuideBridge {
...
+        this.#userConfig[key] = config;
+    }
+
```
nit23uec, line 255: where are we consuming this config?

ui.frontend/src/GuideBridge.js
deepprakash345: Sure, will do this

bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/aemform/AEMFormImpl.java
deepprakash345: Ok, will make this change

ui.frontend/src/GuideBridge.js
deepprakash345: Hi @nit23uec ,
As per our last discussion with you and rishi, we decided to just store the configs for now .

rismehta: Config would be per form, GuideBridge instance is for page and a page can contain multiple forms.

bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/aemform/AEMFormImpl.java
deepprakash345: Hi nitin,
If we expose getVersion then whenever our fd:version is updated all the consuming code of this api needs to make that change but if we expose wrapper methods then that change will be made by us at one place in the model itself.
Isn't that better

bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/aemform/AEMFormImpl.java
deepprakash345: changed to read from model, removed method arguments.

bundles/core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/aemform/AEMFormImpl.java
deepprakash345: exposed getVersion Api

bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/aemform/AEMForm.java
deepprakash345: exposed getVersion , so version check is added.

ui.frontend/src/GuideBridge.js
deepprakash345: added

ui.frontend/src/GuideBridge.js
deepprakash345: Config registered per formPath now

bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/aemform/AEMForm.java
```diff
@@ -352,4 +352,15 @@ default List<HtmlPageItem> getHtmlPageItems() {
         return new ArrayList();
     }
 
+    /**
+     * Indicates if the form configured is v2 adaptive form or not
+     *
+     * @return true if the form is v2 adaptive form; false otherwise
```
nit23uec, line 358: pls correct javadocs.

ui.frontend/src/GuideBridge.js
nit23uec: we should use the strategy we use for other guidebridge APIs. (where formPath is taken from connect initialisation)
formpath should not be passed as arg in any of the other APIs.

nit23uec: Also, till we are not consuming the configs I'd suggest not to expose this API, rather not call this API if form is v2.

bundles/core/src/main/java/com/adobe/cq/forms/core/components/models/aemform/AEMForm.java
deepprakash345: done

ui.frontend/src/GuideBridge.js
deepprakash345: removed the api and added check to skip this api call from form app on the basis of formVersion

ui.frontend/src/GuideBridge.js
```diff
@@ -23,6 +23,8 @@ export default class GuideBridge {
     #guideContainerMap = {};
     #guideBridgeConnectHandlers = [];
     #formContainerPath = "";
+    #userConfig = {};
```
nit23uec, line 26: userConfig should be removed as its unused.

nit23uec (approved):

nit23uec (approved):

rismehta:
Let's create a test case for this later. We should merge this PR since it might block the release of Adaptive Form Embed component

rismehta:
Merging. @deepprakash345 please tag me on the test PR once this is done.

PR #480, opened 2023-02-20
add localisation for thankyouMessage

s1mahanty:
## Related Issue
[FORMS-3187](https://jira.corp.adobe.com/browse/FORMS-3187)

## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractComponentImpl.java
```diff
@@ -84,6 +85,8 @@ public abstract class AbstractComponentImpl implements Component {
      */
     private String id;
 
+    protected I18n i18n = null;
```
nit23uec, line 88: pls move initialisation of i18n in this class from AbstractFormComponentImpl.

rismehta:
We would also need to change the rest API, @ravise5 is doing the same

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractComponentImpl.java
s1mahanty: i18n is initialized in the base class, as discussed offline.

rismehta:
Can you add a small unit test case for localization ?

PR #578, opened 2023-04-14
FORMS-8081: Showing only Adaptive Forms components as options in replace dialog

egmox:
## Types of changes

## Checklist:

### DIFF
--- a/ui.tests/test-module/specs/replaceaction/replacelayout-container.authoring.spec.js
+++ b/ui.tests/test-module/specs/replaceaction/replacelayout-container.authoring.spec.js
@@ -21,10 +21,32 @@ const sitesSelectors = require('../../libs/commons/sitesSelectors'),
  * Testing Form Component replace behaviour in authoring
  */
 describe('component replace - Authoring', function () {
+    const fieldTypes = {TEXT: 'text', SELECT: 'select', NON_INPUT: 'nonInputReadOnly'}
+    const typeMap = {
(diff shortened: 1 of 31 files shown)


ui.tests/test-module/specs/replaceaction/replacelayout-container.authoring.spec.js
```diff
@@ -79,14 +80,14 @@ describe('component replace - Authoring', function () {
         // Check If Dialog Options Are Visible
         if (componentEditPathSelector == buttonEditPathSelector) {
             cy.get(textInput)
-                .should("exist");
+                .should("not.exist");
             cy.get(submitButton)
                 .should("exist");
-            cy.get(textInput)
```
deepprakash345, line 85: @egmox , It seems like we have tested for button and panels , but what about all other input type components and non input type components that we have, can we write a generic test case to check replace behaviour for them also.
cc:- @nit23uec , what do you suggest

deepprakash345 (changes requested):

ui.tests/test-module/specs/replaceaction/replacelayout-container.authoring.spec.js
egmox: @deepprakash345 In the frontend, we don't have any parameters that tells if it is input type or not. We will need to check list of all components if it is showing in dialog or not

PR #582, opened 2023-04-17
added testcase for aemform embed container runtime in sites in iframe and non iframe mode

deepprakash345:


deepprakash345:
@rismehta , @sakshi-arora1 , @nit23uec  cypress test case for aem embed container in iframe mode and non iframe mode

rismehta (approved):

PR #585, opened 2023-04-18
FORMS-3234 Added support to override same values for enum in Dropdown, Check Box Group, and Radio Button component

amanthakur03:
…, Check Box Group, and Radio Button component

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractOptionsFieldImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractOptionsFieldImpl.java
@@ -16,6 +16,7 @@
 package com.adobe.cq.forms.core.components.util;
 
 import java.util.Arrays;
+import java.util.HashMap;
 
 import org.apache.sling.models.annotations.Default;
 import org.apache.sling.models.annotations.injectorspecific.InjectionStrategy;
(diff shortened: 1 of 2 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractOptionsFieldImpl.java
```diff
@@ -51,6 +52,18 @@ public boolean isEnforceEnum() {
         return enforceEnum;
     }
 
+    public String[] removeDuplicates(boolean flag, int length) {
+        String[] enumArray = this.enums;
```
rismehta, line 56: Please make use of streams, code is much more readable.

rismehta, line 56: enums could also be other data types like number, boolean etc. Why are you using string here

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractOptionsFieldImpl.java
```diff
@@ -51,6 +52,18 @@ public boolean isEnforceEnum() {
         return enforceEnum;
     }
 
+    public String[] removeDuplicates(boolean flag, int length) {
```
rismehta, line 55: What do you mean by flag ? what is length ? Create a function which manipulates the map via streams.

rismehta (changes requested):
-1 for now

PR #590, opened 2023-04-18
FORMS-3597 DoR Template configuration

dmaurya929:
## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
@@ -57,6 +57,7 @@ public class FormContainerImpl extends AbstractContainerImpl implements
 
     private static final String DOR_TYPE = "dorType";
     private static final String DOR_TEMPLATE_REF = "dorTemplateRef";
+    private static final String DOR_TEMPLATE_CHANGED = "dorTemplateChanged";
 
     private static final String DOR_TEMPLATE_TYPE = "dorTemplateType";
     private static final String FD_SCHEMA_TYPE = "fd:schemaType";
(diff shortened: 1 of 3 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -57,6 +57,7 @@ public class FormContainerImpl extends AbstractContainerImpl implements
 
     private static final String DOR_TYPE = "dorType";
     private static final String DOR_TEMPLATE_REF = "dorTemplateRef";
+    private static final String DOR_TEMPLATE_CHANGED = "dorTemplateChanged";
```
deepprakash345, line 60: where is this used?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -260,4 +264,22 @@ public Map<String, Object> getDorProperties() {
         }
         return customDorProperties;
     }
+
+    @Override
+    @JsonIgnore
+    public String getDorType() {
```
deepprakash345, line 270: we already have a getter i suppose, getDorProperties which returns these values, why do we need these separate getters. we should have only one

deepprakash345 (changes requested):

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
dmaurya929: This property is used to show message if acroForm templates is changed and review the bindings
Currenlty this properties is saved in asset node of form, but going forward with sites  we want this to persist in form conatiner node. So exposing API to read this property.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
dmaurya929: dorType and dorTemplateRef is specific to form container, I saw **getDorProperties** return a map depending upon component. Isn't it will be good to return specific property as we do for schemaRef and schemaType ?

PR #641, opened 2023-05-16
[FORMS-7963] added support for hiding nav bars in layout components i…

deepprakash345:
added support for hiding nav bars in layout components if child is hidden along with test cases and caveats handling

- Tabs
-  Wizard (Handled navigation as well)
-  Accordion

Added test cases for rule based hiding and jcr based hiding.


barshat7:
Test cases for vertical tabs are missing. Is it intentional?

PR #658, opened 2023-05-23
FORMS-8863 containsFormContainer API to check if resource contains ad…

barshat7:
…aptive form container

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
@@ -43,6 +43,26 @@ public class FormStructureParserImpl implements FormStructureParser {
         return getPropertyFromFormContainer(resource, FormContainer.PN_CLIENT_LIB_REF);
     }
 
+    @Override
+    public Boolean containsFormContainer() {
+        return containsFormContainer(resource);
+    }
+
+    private Boolean containsFormContainer(Resource resource) {
+        if (resource == null) {
+            return false;
+        }
+        if (ComponentUtils.isAFContainer(resource)) {
+            return true;
+        }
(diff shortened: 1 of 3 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormStructureParser.java
```diff
@@ -37,4 +37,9 @@ public interface FormStructureParser {
      */
     String getClientLibRefFromFormContainer();
 
+    /**
+     *
+     * @return true if this resource contains a form container
+     */
+    Boolean containsFormContainer();
```
rismehta, line 44: Can you improve the function documentation ?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
```diff
@@ -43,6 +43,26 @@ public String getClientLibRefFromFormContainer() {
         return getPropertyFromFormContainer(resource, FormContainer.PN_CLIENT_LIB_REF);
     }
 
+    @Override
+    public Boolean containsFormContainer() {
+        return containsFormContainer(resource);
+    }
+
+    private Boolean containsFormContainer(Resource resource) {
```
rismehta, line 51: Why did you create another method, it is confusing. Can you add it the original method ?

rismehta (changes requested):
-1 for now

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
barshat7: We'll need to recursively traverse the children nodes to check whether there exists a resource with fieldType is form. Its a private recursive method.

PR #685, opened 2023-06-05
FORMS-3225 Added fix for blank title getting replaced with component's name

amanthakur03:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
```diff
@@ -78,6 +78,10 @@ public Boolean isVisible() {
      */
     @Override
     public String getValue() {
-        return ComponentUtils.translate(properties.get(PN_TITLE, this.defaultTitle), PN_TITLE, properties, i18n);
+        if (properties.containsKey(PN_TITLE))
```
rismehta, line 81: If PN_TITLE is not present in properties, we always return the default title, which is the name of the text box. This behavior is as per AF v1. Why do we need to change this ?

rismehta (changes requested):
-1 for now. use-case is not clear

bundles/af-core/src/test/resources/form/checkboxgroup/test-content.json
```diff
@@ -52,7 +52,7 @@
     "id": "checkboxgroup-40e63c6f46",
     "sling:resourceType": "core/fd/components/form/checkboxgroup/v1/checkboxgroup",
     "fieldType": "checkbox-group",
-    "name": "Personality",
+    "name": "",
```
rismehta, line 55: name can never be blank as per af specification.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
amanthakur03: @rismehta I see in AF1 we have the behavior as the component's title remains blank if given blank. We are trying to implement the same here. Not sure if i am missing anything. Pls suggest..??
image

bundles/af-core/src/test/resources/form/textinput/test-content.json
```diff
@@ -93,7 +93,7 @@
     "readOnly": false,
     "autoComplete": "off",
     "label": {
-      "value": "Full Name"
```
ujjwal5, line 96: please add new collateral instead of changing existing collaterals. Applicable elsewhere as well.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
```diff
@@ -78,6 +78,10 @@ public Boolean isVisible() {
...
+            return ComponentUtils.translate(properties.get(PN_TITLE, this.defaultTitle), PN_TITLE, properties, i18n);
```
ujjwal5, line 82: if we are checking if PN_TITLE exists, why send default value as argument to function ?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
```diff
@@ -78,6 +78,10 @@ public Boolean isVisible() {
...
+        else
+            return ComponentUtils.translate(properties.get(PN_TITLE, ""), PN_TITLE, properties, i18n);
```
ujjwal5, line 84: Please see if we are saving "" in AF v1 or "\n".
Also check if the same is being reflected on content tree as well?

ujjwal5 (changes requested):

bundles/af-core/src/test/resources/form/textinput/test-content.json
amanthakur03: @ujjwal5 I have added new collateral as well but still we need to make changes in existing collateral since this collateral contains values according to the earlier behavior of AF v2 (Title gets overridden with name of the component on providing blank input). These test cases start to fail if the values don't get changed in collateral.

sakshi-arora1 (commented):
Don't change existing tests, create anew collateral with all components and empty labels.

ui.tests/pom.xml
```diff
@@ -24,7 +24,7 @@
     <parent>
         <groupId>com.adobe.aem</groupId>
         <artifactId>core-forms-components-parent</artifactId>
-        <version>2.0.100-SNAPSHOT</version>
+        <version>2.0.19-SNAPSHOT</version>
```
rismehta, line 27: There is no point changing this version. Please remove the other changes and keep this PR only for changes mentioned in the PR description.

rismehta:
Why does this PR have so many changes. Closing it, please open a new one with only relevant changes

PR #696, opened 2023-06-08
[FORMS-7828] resolved problem of navigation not working in authoring without refresh in case of horizontal and vertical tabs

deepprakash345:
Added test case to check this behaviour :-

The issue was because the editor hook sends a navigation event as soon as the element is added to tab before the tab's constructor call . hence the _self._elements cache doesn't have the newly added elements and while checking the array at the index , the call fails and constructor call doesn't happens.

When we refresh the constructor call happens resulting in correct cache formation and hence correct navigation.


rismehta:
Can you please check this flaky test, https://app.circleci.com/pipelines/github/adobe/aem-core-forms-components/3166/workflows/ad5c1dea-fb29-4b32-9bb6-110be6e4090d/jobs/10612/tests. Looks like in site editor, due to scroll bar, the test fails. Please fix the same

rismehta (approved):

PR #699, opened 2023-06-09
FORMS-3225 Fix for blank title getting replaced with component's name

amanthakur03:


barshat7:
Pipeline has not executed. Kindly create a branch from adobe:dev instead of forking it.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
```diff
@@ -78,6 +77,6 @@ public Boolean isVisible() {
      */
     @Override
     public String getValue() {
-        return ComponentUtils.translate(properties.get(PN_TITLE, this.defaultTitle), PN_TITLE, properties, i18n);
+        return ComponentUtils.translate(properties.get(PN_TITLE, "\n"), PN_TITLE, properties, i18n);
```
barshat7, line 80: why do we have to use newline character '\n' instead of empty string as default?

amanthakur03: @barshat7 In order to achieve the same outcome as in AF V1 and as per the comments of @ujjwal5 , we have opted to use line breaks ("\n") instead of an empty string ("").

egmox:
Is it possible to add authoring test also?

amanthakur03:
Closing this PR. Will create a new PR from adobe:dev

PR #701, opened 2023-06-12
FORMS-3225 Fix for blank title getting replaced with component's name

amanthakur03:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
```diff
@@ -78,6 +76,6 @@ public Boolean isVisible() {
      */
     @Override
     public String getValue() {
-        return ComponentUtils.translate(properties.get(PN_TITLE, this.defaultTitle), PN_TITLE, properties, i18n);
+        return ComponentUtils.translate(properties.get(PN_TITLE, "\n"), PN_TITLE, properties, i18n);
```
rismehta, line 79: If there is no title in JCR, why are you returning an empty line ?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
```diff
@@ -31,12 +31,10 @@ public class LabelImpl implements Label {
     private static final String PN_IS_TITLE_RICH_TEXT = "isTitleRichText";
 
     private ValueMap properties;
-    private String defaultTitle;
     private I18n i18n;
 
-    public LabelImpl(Resource field, String defaultTitle, I18n i18n) {
```
rismehta, line 37: What will be the default title if you remove this ?

bundles/af-core/src/test/resources/form/numberinput/test-content.json
```diff
@@ -36,6 +36,7 @@
     "fieldType": "number-input",
     "description": "Enter your phone number.",
     "name": "Phone Number",
+    "jcr:title" : "Phone Number",
```
rismehta, line 39: Why have you added this ?

rismehta (changes requested):
Check comments

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
amanthakur03: @rismehta In order to make it consistent with AF v1 and as per the suggestions of @ujjwal5 we have returned an empty line instead of empty string. I can even change it to empty string too if required. Pls suggest.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/LabelImpl.java
amanthakur03: Here we were using getName() method as default title. Since it was not required as we are checking if jcr:title is present i have removed the reference.

bundles/af-core/src/test/resources/form/numberinput/test-content.json
amanthakur03: As discussed in scrum we have added jcr:title  in order to pass the tests without modifying any values in existing collateral.

PR #712, opened 2023-06-20
Forms 9357  added cleanTitle test command which deletes residual title components from previous tests, separate command needed because we can't delete by exact editPath in case of title component, we delete by matching startsWith in string...

deepprakash345:
Solves flaky tests in cases where residual title component is left behind in the test form from some previously failed test.

### DIFF
--- a/ui.tests/test-module/libs/support/commands.js
+++ b/ui.tests/test-module/libs/support/commands.js
@@ -41,7 +41,7 @@
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
 
 import 'cypress-file-upload';
-import { recurse } from 'cypress-recurse'
+import {recurse} from 'cypress-recurse'
 
 const commons = require('../commons/commons'),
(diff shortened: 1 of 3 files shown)


ui.tests/test-module/libs/support/commands.js
```diff
@@ -330,6 +332,22 @@ Cypress.Commands.add("cleanTest", (editPath) => {
     });
 })
 
+Cypress.Commands.add("cleanTitleTest", (editPath) => {
```
nit23uec, line 335: can we have something reusable for all components if required. something like `removeComponents`
Also ,it's not being used in sites editor test. any reasons? we should be consistent about its usage here  otherwise it may confuse the developers in future whether to include it or not everytime.

deepprakash345: Hi @nit23uec , we already have a generic command named cleanTest. Will add this in the site test as well

PR #728, opened 2023-07-03
FORMS-8306 Added an enhancement support to display 'Please drag {Panel Title} components here' instead of the generic 'Drag components here' for panel component."

ghost:


ui.tests/test-module/specs/panelcontainer/panelcontainer.authoring.spec.js
```diff
@@ -94,6 +103,34 @@ describe('Page - Authoring', function () {
...
+            dropPanelInContainer();
+            cy.openEditableToolbar(sitesSelectors.overlays.overlay.component + panelContainerPathSelector);
+            cy.invokeEditableAction("[data-action='CONFIGURE']");
+            cy.get('input[name="./name"]').invoke('val', 'customPanel');
+            cy.get('input[name="./jcr:title"]').invoke('val', 'Custom Panel');
+            cy.get('.cq-dialog-submit').click().then(() => {
+                cy.get('.cq-dialog-submit').should('not.exist');
+            });
+            getPreviewIframeBody().find('.cmp-container__label').should('have.text', 'Custom Panel')
+            getPreviewIframeBody().find("div[data-cmp-adaptiveformcontainer-path='/content/forms/af/core-components-it/blank/jcr:content/guideContainer']").should('have.attr', 'data-placeholder-text', 'Please drag Custom Panel components here');
```
egmox, line 118: why do we need to call getPreviewIframeBody() function for both lines 117 and 118? Can't we just use cy.get() ?

ui.tests/test-module/specs/panelcontainer/panelcontainer.authoring.spec.js
```diff
@@ -94,6 +103,34 @@ describe('Page - Authoring', function () {
...
+            dropPanelInContainer();
+            cy.openEditableToolbar(sitesSelectors.overlays.overlay.component + panelContainerPathSelector);
+            cy.invokeEditableAction("[data-action='CONFIGURE']");
+            cy.get('input[name="./name"]').invoke('val', 'customPanel');
+            cy.get('input[name="./jcr:title"]').invoke('val', '');
+            cy.get('.cq-dialog-submit').click().then(() => {
+                cy.get('.cq-dialog-submit').should('not.exist');
+            });
+            getPreviewIframeBody().find('.cmp-container__label').should('not.exist')
+            getPreviewIframeBody().find("div[data-cmp-adaptiveformcontainer-path='/content/forms/af/core-components-it/blank/jcr:content/guideContainer']").should('have.attr', 'data-placeholder-text', 'Drag Panel components here');
```
egmox, line 132: Same as above? Can't we just use cy.get() ?

amanthakur03: In order to access the elements which is present in Content Frame we need to call getPreviewIframeBody() method else the elements won't be visible.

ui.tests/test-module/specs/panelcontainer/panelcontainer.authoring.spec.js
amanthakur03: Same reason as below.

ui.tests/test-module/specs/panelcontainer/panelcontainer.authoring.spec.js
rismehta, line 132: Can you re-use the constant defined above to access form container path ?

rismehta (changes requested):
-1 for now. Can you test the localization aspect, and attach the screenshots in this PR.

ui.tests/test-module/specs/panelcontainer/panelcontainer.authoring.spec.js
amanthakur03: Sure

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/panelcontainer/v1/panelcontainer/responsiveGrid.html
```diff
@@ -22,6 +22,8 @@
          data-cmp-adaptiveformcontainer-path="${formstructparser.formContainerPath}"
          title="${panel.tooltipVisible ? '' : panel.tooltipText}"
          role="${container.roleAttribute}"
+         data-sly-test.placeholderText="${'Please drag {0} components here' @ i18n, format=[panel.label.value]}"
+         data-placeholder-text="${panel.label.value ? placeholderText : 'Drag Panel components here'}"
```
rismehta, line 26: How will localization work for dynamic labels ? We would have to document this.

Both the strings have to be `Granite.I18n.get`

Can you make use of

rismehta (changes requested):
-1 for now

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/panelcontainer/v1/panelcontainer/responsiveGrid.html
nit23uec: label is generally according to the type of form field from an user persistence.
if the label of the panel is "Enter you address details", then would we say: 
"Please drag Enter you address details components here"
I don't think it would make sense.
I'd still prefer to use "'Drag Panel components here" irrespective of the label.

deepprakash345: @amanthakur03 , i think we need to discuss the case nitin is referring to before merging this

deepprakash345 (changes requested):

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/panelcontainer/v1/panelcontainer/responsiveGrid.html
amanthakur03: Yes Deep, Posting on slack for everyone's suggestions.

rismehta: > I'd still prefer to use "'Drag Panel components here" irrespective of the label.

Should we rather use name here ? The problem with name is that, sometimes business author don't change the auto generated name, unless they use it in rules.

vdua: > I'd still prefer to use "'Drag Panel components here" irrespective of the label.
That gets confusing when you have nested hierarchy. `Drag Panel components(Title) here` or on hover show the title.

PR #735, opened 2023-07-10
FORMS-9364 Edited the code for calling name for panel component

AnurudraS:
## Types of changes

## Checklist:

### DIFF
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/panelcontainer/v1/panelcontainer/responsiveGrid.html
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/panelcontainer/v1/panelcontainer/responsiveGrid.html
@@ -16,6 +16,7 @@
 <template data-sly-template.responsiveGrid="${ @ container, panel, label, shortDescription, longDescription, questionMark}">
     <div data-sly-use.formstructparser="com.adobe.cq.forms.core.components.models.form.FormStructureParser"
          id="${panel.id}"
+         name="${panel.name}"
          class="cmp-container"
          data-cmp-is="adaptiveFormPanel"
          data-cmp-data-layer="${panel.data.json}"


ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/panelcontainer/v1/panelcontainer/responsiveGrid.html
```diff
@@ -16,6 +16,7 @@
 <template data-sly-template.responsiveGrid="${ @ container, panel, label, shortDescription, longDescription, questionMark}">
     <div data-sly-use.formstructparser="com.adobe.cq.forms.core.components.models.form.FormStructureParser"
          id="${panel.id}"
+         name="${panel.name}"
```
rismehta, line 19: The <div> element is a generic container in HTML and does not have any inherent form-related functionality. Therefore, the name attribute is not applicable or meaningful for <div> elements in the context of HTML.

rismehta (changes requested):
Please check comments

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/panelcontainer/v1/panelcontainer/responsiveGrid.html
dmaurya929: I think the intent is to call getter `getName()`, it checks if `name` property exists else it sets this property.
When we insert panel component, on rendering of this component `getName` doesn't get called. So in edit dialog name field is empty until we refresh the page.

dmaurya929: We should this problem in model itself ( may be init of model) instead of HTL. As if any component will extend panel container then that may also face the same issue

AnurudraS:
review comments are incorporated in below request:

https://github.com/adobe/aem-core-forms-components/pull/737

PR #751, opened 2023-07-17
Forms 8533 replace comp with same rte

egmox:
## Types of changes

## Checklist:


examples/ui.content/src/main/content/jcr_root/conf/core-components-examples/settings/wcm/policies/.content.xml
```diff
@@ -142,7 +142,7 @@
                             jcr:primaryType="nt:unstructured"
                             jcr:title="formcontainer-policy"
                             sling:resourceType="wcm/core/components/policy/policy"
-                            components="[group:Core Components Examples - Adaptive Form,group:Core Components Examples - Forms And Communications Portal,group:Core Components Examples]">
+                            components="[group:Core Components Examples - Adaptive Form,group:Core Components Examples - Forms And Communications Portal,group:Core Components Examples,group:replace test group]">
```
rismehta, line 145: Please don't change examples for test case, ideally any test case related changes should be in the collateral present as part of `it` folder, since examples is delivered to aemcomponents.dev

rismehta (changes requested):
-1 for now

it/apps/src/main/content/META-INF/vault/filter.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <workspaceFilter version="1.0">
     <filter root="/apps/forms-core-components-it"/>
+    <filter root="/apps/forms-core-components-it/form/button"/>
```
rismehta, line 4: We already have the root filter why do you need to add this ?

it/content/src/main/content/jcr_root/content/dam/formsanddocuments/core-components-it/samples/imagetestgroup/.content.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rismehta, line 1: Inside samples create a replace folder, inside replace folder create collateral based on the use-case, for example, in this case, the name of the form would be `replacewcmcomponents`

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/replacehook.js
```diff
@@ -131,26 +131,23 @@
 
                 if (!(keyword.length > 0) || isKeywordFound) {
                     if ((!cannotBeReplacedWith.includes(componentType))
-                        && typeMap[editableType] === typeMap[componentType]
```
rismehta, line 134: I would want to move this code to cq-guides, can you create a tracking ticket for this. To me, this logic should not be part of core component, this is not something which forms developers would like to customize ?

rismehta (changes requested):
-1 for now

examples/ui.content/src/main/content/jcr_root/conf/core-components-examples/settings/wcm/policies/.content.xml
egmox: Would it be fine if using cypress, we just update the template checking the 'replace test group'. This way, it will not be in code.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/replacehook.js
rismehta: > This is just an implementation detail for replace hook that may /may not be customised but it doesn't matters to us.

Sorry I beg to differ here, I would want to keep the core component code clean. We have some API's already exposed in cq-guides which is consumed by core components. Core component should only contain code which is meaningful for forms developer to read, else it adds complexity. Replace is an edit action and forms core component developer should just use it (as API's)

rismehta (approved):

PR #760, opened 2023-07-19
FORMS-9364 Changes for the name in panel edit.

AnurudraS:
## Description
Changes for Missing name in panel.

## Related Issue
[FORMS-9364](https://jira.corp.adobe.com/browse/FORMS-9364) -  Name field is showing empty for Panel component.

## Types of changes

## Checklist:


bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImplTest.java
```diff
@@ -305,4 +305,11 @@ private FormContainer getFormContainerWithLocaleUnderTest(String resourcePath) t
         context.currentResource().adaptTo(FormContainer.class);
         return request.adaptTo(FormContainer.class);
     }
+
+    @Test
+    void testGetName() throws Exception {
+        FormContainer formContainer = Utils.getComponentUnderTest(PATH_FORM_1, FormContainer.class, context);
+        assertNull(formContainer.getName());
+        // assertEquals("L2NvbnRlbnQvZm9ybXMvYWYvZGVtbw==", formContainer.getId());
```
s1mahanty, line 313: you may remove the unwanted commented code, please.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -292,4 +292,12 @@ public String getParentPagePath() {
         }
         return StringUtils.EMPTY;
     }
+
+    @JsonIgnore
+    @Override
+    public String getName() {
```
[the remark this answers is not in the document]
AnurudraS: > 

yes your review make sense but here in our scenario in abstractFormComponentImpl.java we have override it with the definition. Therefore if we don't write in formContainerIMPL then the name will be set every time for both FormContainer and element. Now the name for element(panel) will be set and formContainer will be set to null.

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImplTest.java
AnurudraS: > you may remove the unwanted commented code, please.

done!

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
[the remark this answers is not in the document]
s1mahanty: But do we need to do it twice in FormContainer and it's Impl, when both places the function definition is the same ?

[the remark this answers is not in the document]
s1mahanty: I mean it is redundant refer: https://github.com/adobe/aem-core-forms-components/blob/master/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormContainer.java#L162

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -116,6 +116,7 @@ protected void initBaseModel() {
         if (Boolean.TRUE.equals(unboundFormElement)) {
             dataRef = NULL_DATA_REF;
         }
+        getName();
```
rismehta, line 119: The issue is only for panel, can we fix it only for panel ?

rismehta (changes requested):
minor comments

rismehta:
@AnurudraS There are test case failure due to your changes.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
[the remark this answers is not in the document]
deepprakash345: > s [FormContainer] that i

Depends on the use cases, as the name suggests default methods provide the default implementation but different child classes can decide to override it or implement their own version of same method. Those who don't need it to be different than the parent class need not ovverride it .

In the current scenario, I don't think there is any need to re-implement it. cc:- @s1mahanty , @AnurudraS

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -155,12 +156,12 @@ public String getText() {
...
     public String getName() {
         if (name == null) {
-            name = getDefaultName();
+            name = getNameOrSetDefault();
         }
         return name;
     }
 
-    protected String getDefaultName() {
+    protected String getNameOrSetDefault() {
```
deepprakash345, line 164: @AnurudraS ,do name the method appropriately, by the name it looks like this method will check if name is not null then return the current value of null otherwise will set the name and then return. Either change the name of move the null check from getName to here.
Why do you need a separate method, why can't you do this in the getName method only if name is null.

deepprakash345 (changes requested):
-1 for now

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -116,7 +116,7 @@ protected void initBaseModel() {
         if (Boolean.TRUE.equals(unboundFormElement)) {
             dataRef = NULL_DATA_REF;
         }
-        getNameOrSetDefault();
+        getName();
```
AnurudraS, line 119: function name changes because getNameOrSetDefault was setting default value for others as well. Other test cases were failing

rismehta:
Can you reformat the file ? Please run the build command before raising a PR.

AnurudraS:
> Can you reformat the file ? Please run the build command before raising a PR.

Screenshot 2023-08-02 at 3 17 10 PM

Build was success in local checking and correcting

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
AnurudraS: > 
On removing the name the code is lossing the code coverage because we cannot write test case for name = null as the resource cannot be null or the Key name cannot be set to null. However the method is kept protected so it should not break or impact the other properties.

PR #788, opened 2023-08-02
FORMS-10107 Fix id mismatch issue with embed form having fragment

dmaurya929:
## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractComponentImpl.java
```diff
@@ -129,7 +129,8 @@ private void init() {
         if (currentPage != null && resource != null && request.getAttribute(REQ_ATTR_EMBEDDED_ADAPTIVEFORM) != null) {
             if (!GuideWCMUtils.isForms(getCurrentPage().getPath())) {
                 PageManager pageManager = currentPage.getPageManager();
-                Page resourcePage = pageManager.getContainingPage(resource);
+                String pagePath = (String) request.getAttribute(REQ_ATTR_EMBEDDED_ADAPTIVEFORM);
```
rismehta, line 132: What if is this attribute is not set ?

it/content/src/main/content/META-INF/vault/filter.xml
```diff
@@ -8,4 +8,5 @@
     <filter root="/conf/core-components-it" mode="update"/>
     <filter root="/conf/wknd" mode="update"/>
     <filter root="/conf/datalayerContainer" mode="update"/>
+    <filter root="/content/core-components-examples/aemembedcontainertest"/>
```
rismehta, line 11: All sites specific use-case should be present under `/content/forms/sites`. No changes in existing filter is allowed unless there are new use-cases

rismehta (changes requested):
-1 for now.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractComponentImpl.java
dmaurya929: In line 129 there is already a condition in place to check if this attribute is not null. Cypress test covers this scenario, let me see about the unit test.

PR #804, opened 2023-08-07
FORMS-9757 avoid re-declaration of widget classes in case of loading …

barshat7:
…different form

## Types of changes

## Checklist:


rismehta (changes requested):
Please add cypress test cases

it/apps/src/main/content/jcr_root/apps/forms-core-components-it/samplepages/loadunloadaf/v1/loadunloadaf/loadunloadaf.html
```diff
@@ -0,0 +1,43 @@
...
+           function loadform(btn) {
+             var formPath = btn.getAttribute('data-form-path')
+             loadAdaptiveForm({
+               formPath: formPath,
+               CSS_Selector: '.customafsection'
+             })
+           }
+
+           var loadAdaptiveForm = function(options) {
+             var path = options.formPath + '.html?wcmmode=disabled';
```
rismehta, line 13: Can you write a comment that on publish instance the query parameters are not required

it/apps/src/main/content/jcr_root/apps/forms-core-components-it/samplepages/loadunloadaf/v1/loadunloadaf/loadunloadaf.html
```diff
@@ -0,0 +1,43 @@
...
+             fetch(path).then((res) => res.text()).then(html => {
+               setInnerHTML(document.querySelector(options.CSS_Selector), html); // OR $(options.CSS_Selector).html(html);
+             })
+           }
+
+            function setInnerHTML(elm, html) {
+                elm.innerHTML = html;
+                Array.from(elm.querySelectorAll("script"))
```
rismehta, line 21: Why do you need to set script tag individually ? innerHTML should do it

it/apps/src/main/content/jcr_root/apps/forms-core-components-it/samplepages/loadunloadaf/v1/loadunloadaf/loadunloadaf.html
rismehta, line 13: how will this work if one needs to only show the form container (ie) without header/footer. In those case, we would have to also include the runtime library in the main page. Can you write a test case for this and also update the documentation ?

rismehta (changes requested):
Check comments

it/apps/src/main/content/jcr_root/apps/forms-core-components-it/samplepages/loadunloadaf/v1/loadunloadaf/loadunloadaf.html
barshat7: script set through innerHTML does not execute automatically.
https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#:~:text=in%20this%20case-,Although%20this,-may%20look%20like
We can either use jquery or if jquery is not present then the given method is one of the ways to include scripts in embedded html.

PR #807, opened 2023-08-08
FORMS-10186 Fixing GuideBridge events

rismehta:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/constants.js
+++ b/ui.frontend/src/constants.js
@@ -84,6 +84,129 @@ export const Constants = {
     FORM_LANGUAGE_INITIALIZED: "AF_LanguageInitialised",
 
     /**
+     * Event triggered when focus is changed on a form element
+     * @event module:FormView~Constants#ELEMENT_FOCUS_CHANGED
+     * @property {object} event - The event object.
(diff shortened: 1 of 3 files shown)


ui.frontend/src/constants.js
```diff
@@ -83,6 +83,129 @@ export const Constants = {
      */
     FORM_LANGUAGE_INITIALIZED: "AF_LanguageInitialised",
 
+    /**
+     * Event triggered when focus is changed on a form element
+     * @event module:FormView~Constants#ELEMENT_FOCUS_CHANGED
```
nit23uec, line 88: The documentation should be there on the event itself in GuideBridge for easy discoverability.

rismehta: What do you mean by discoverability ? Can you explain. All the events are triggered from view layer and we don't document the view layer code today (only ui.frontend is documented). 

IMO, only thing missing in the documentation is the reference from where the event is thrown (ie) @fires annotation, an example can be seen here, https://opensource.adobe.com/aem-core-forms-components/module-FormView-Utils.html#.setupFormContainer__anchor

PR #813, opened 2023-08-09
ID mismatch with experience fragment

dmaurya929:
## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractContainerImpl.java
```diff
@@ -105,6 +109,17 @@ protected void initBaseModel() {
         if (request != null && resource.isResourceType("wcm/foundation/components/responsivegrid")) {
             resGrid = request.adaptTo(ResponsiveGrid.class);
         }
+
+        if (request != null) {
```
sakshi-arora1, line 113: This logic will execute for all containers, can we make it specific to form container. You can add one more layer of abstraction specific to form container

dmaurya929: Ok, it makes sense. This logic is specific to form container so will add it there only.

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImplTest.java
```diff
@@ -111,7 +111,7 @@ public void testEmbedWithIframe() {
         Mockito.when(sitePage.getPath()).thenReturn(PAGE_PATH);
         Mockito.when(sitePage.getPageManager()).thenReturn(pageManager);
         MockSlingHttpServletRequest request = context.request();
-        request.setAttribute("embeddedAdaptiveForm", AF_PATH);
+        request.setAttribute("referencedPage", AF_PATH);
```
rismehta, line 114: Can you re-use the constant in the test too ?

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/container.html
```diff
@@ -39,5 +39,6 @@
              data-sly-resource="${paragraph.path @ resourceType=paragraph.resourceType, decorationTagName='div', cssClassName=paragraph.cssClass}"></sly>
         <sly data-sly-resource="${resource.path @ resourceType='wcm/foundation/components/responsivegrid/new', appendPath='/*', decorationTagName='div', cssClassName='new section aem-Grid-newComponent'}"
              data-sly-test="${!isAllowedApplicable && (wcmmode.edit || wcmmode.preview)}"></sly>
+        <sly data-sly-use="${'removeattribute.js' @ referencedPage }"/>
```
rismehta, line 42: Can you re-use the constant defined in JAVA class here ? It helps to understand the usage during search

rismehta (changes requested):
check comments

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/container.html
dmaurya929: FormConstants is present in package `com.adobe.cq.forms.core.components.internal`, here we have keyword `internal` that's why not using in htl file. Please let me know if still it should be added.

sakshi-arora1 (approved):

rismehta:
Please fix the conflicts and re-create the PR

dmaurya929:
@sakshi-arora1 I made the changes as you suggested to set the request attribute of clientLibs for fragments.
However instead of loading these clientlibs in `customfooterlibs` ( as discussed earlier)  I'm loading in the `container.html` itself because in sites we don't load cusomfooterlibs.
Also to avoid duplication load in case of multiple container I have used `Map<String, Boolean>` instead of `Set`, it will keep track if a particular library is already loaded or not.

sakshi-arora1 (commented):
discussed offline with @dmaurya929. Will be making changes.

sakshi-arora1 (approved):

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormStructureParser.java
```diff
@@ -43,4 +47,20 @@ public interface FormStructureParser {
      * @return true if this resource or one of its children is a form container, else false
      */
     Boolean containsFormContainer();
+
+    /**
+     * Adds a client lib reference.
+     * 
+     * @param clientLibRef
+     */
+    default void addClientLibRef(@NotNull String clientLibRef) {}
```
rismehta, line 56: Can you improve the java doc, it is not clear as to what this API is doing.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormStructureParser.java
```diff
@@ -43,4 +47,20 @@ public interface FormStructureParser {
      * @return true if this resource or one of its children is a form container, else false
      */
     Boolean containsFormContainer();
+
```
rismehta, line 50: Since addClientLibRef doesn't directly relate to the core responsibilities of parsing form structures, it would be better to separate this functionality into a different component. 

Let's do the following,
```
public interface FormClientLibManager {
    void addClientLibRef(String clientLibRef);
    List<String> getClientLibReferences()
}
``
We should instead use this interface to manage all forms client library references rather than polluting forms structure parser whose core responsibility is to just parse form structure.

Comment (rismehta) [reply bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormStructureParser.java:50]: Please add proper java docs for the new interface, update the package-info.java file. Not sure why the maven plugin isn't complaining to bump the version

**Review by rismehta — CHANGES_REQUESTED** [review]
-1 for now

File: bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormClientLibManagerImpl.java [hunk 6]
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.sling.models.annotations.Model;
+import org.apache.sling.models.annotations.injectorspecific.InjectionStrategy;
+import org.apache.sling.models.annotations.injectorspecific.SlingObject;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import com.adobe.cq.forms.core.components.models.form.FormClientLibManager;
+
+@Model(
+    adaptables = { SlingHttpServletRequest.class, Resource.class },
```
sakshi-arora1, line 34: why is it adaptable to resource?

dmaurya929: Not required right now, but in future if we will clean form structure parser and move clientlib related code here then it will be required.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormClientLibManager.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Defines the {@code FormClientLibManager} Sling Model to perform operation related to client libraries.
+ *
+ * @since com.adobe.cq.forms.core.components.models.form 4.5.0
+ */
+public interface FormClientLibManager {
+
+    /**
+     * Adds a client lib reference associated with a fragment/form container to a global store in a particular request.
```
sakshi-arora1, line 30: How about:
`Adds a client library associated with form or fragment for a request, that can be consumed later within the scope of request.`
@rismehta do you have a better description?
@dmaurya929 Just want to avoid use of global store

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormClientLibManager.java
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @param clientLibRef
+     */
+    void addClientLibRef(@NotNull String clientLibRef);
+
+    /**
+     * Returns a unique list of client lib references that was added to the global store in a particular request.
```
sakshi-arora1, line 37: `Returns list of client libraries associated with form and fragments that were added for the request.`

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormClientLibManagerImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        if (clientLibSet == null) {
+            clientLibSet = new HashSet<>();
+            request.setAttribute(FormConstants.REQ_ATTR_CLIENT_LIBS, clientLibSet);
+        }
+        clientLibSet.add(clientLibRef);
+    }
+
+    @Override
+    public List<String> getClientLibRefList() {
+        if (request != null && request.getAttribute(FormConstants.REQ_ATTR_CLIENT_LIBS) != null) {
```
sakshi-arora1, line 57: This constant should be local to this class.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormClientLibManager.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.List;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Defines the {@code FormClientLibManager} Sling Model to perform operation related to client libraries.
+ *
+ * @since com.adobe.cq.forms.core.components.models.form 4.5.0
+ */
+public interface FormClientLibManager {
```
rismehta, line 27: The interface should be of ProviderType which means this cannot another implementation apart from the one present in the product. please change this

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormClientLibManagerImpl.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.sling.models.annotations.injectorspecific.SlingObject;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import com.adobe.cq.forms.core.components.models.form.FormClientLibManager;
+
+@Model(
+    adaptables = { SlingHttpServletRequest.class, Resource.class },
+    adapters = FormClientLibManager.class)
+public class FormClientLibManagerImpl implements FormClientLibManager {
```
rismehta, line 36: I don't think this can be ever adapted to resource, please change this.

rismehta (changes requested):
check comments

rismehta:
Review comments would be taken interactively by @sakshi-arora1 / @dmaurya929

PR #846, opened 2023-08-30
@releng adding strict mode

rismehta:
## Types of changes

## Checklist:

### DIFF
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/replacehook.js
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/replacehook.js
@@ -13,9 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  ******************************************************************************/
-'use strict';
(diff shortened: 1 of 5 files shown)


ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/replacehook.js
```diff
@@ -13,9 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  ******************************************************************************/
-'use strict';
 (function (window, author, Coral, channel) {
-
+    "use strict";
```
iamsudhanshu, line 17: @rismehta The original code - was it enabling strict mode for all the clientlib code loaded after this?
If so - good catch, thanks for fixing

rismehta: Workflow editor code was enabling this config, hence it was failing there.

Ideally, tools like eslint or webpack enforces this by default at project level, whereas in case of aem client library we need to manually set this.

PR #623, opened 2023-04-27
Rte title

im-shiv:

PR #888, opened 2023-09-25
FORMS-3696 adding rich text support for labels

im-shiv:


im-shiv:
old PR and discussion for reference: https://github.com/adobe/aem-core-forms-components/pull/623

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractOptionsFieldImpl.java
```diff
@@ -95,15 +100,23 @@ public Object[] getEnums() {
     }
 
     @Override
-    public String[] getEnumNames() {
+    public RichText[] getEnumNames() {
```
rismehta, line 103: You just can't change public API syntax, please mark this as deprecated and introduce a new API. This would break existing customer forms.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/OptionsConstraint.java
```diff
@@ -55,7 +55,7 @@ default boolean isEnforceEnum() {
      * The length of enum and enumNames array must match
      *
      * @return the list of enum names
-     * @since com.adobe.cq.forms.core.components.models.form 0.0.1
+     * @since com.adobe.cq.forms.core.components.models.form 5.0.0
      */
-    String[] getEnumNames();
+    RichText[] getEnumNames();
```
rismehta, line 60: This can't be changed since it already exposed, mark it as deprecate and introduce a new function. Please mention in the new function, that this is introduced since which version based on package-info and only bump the major version

rismehta (changes requested):
Kindly rebase your pull request (PR) with an appropriate commit history to ensure the successful execution of the sync PR GitHub action.

PR #914, opened 2023-10-09
embed dependency as part of all.zip so that it does not conflict with…

barshat7:
… archetype project

## Types of changes

## Checklist:


rismehta (changes requested):
Please write test case for the same

barshat7:
> Please write test case for the same

Test cases were there but the example was also pointing to wcm foundation that's why it was not able to detect. Updated the examples component also so that test is executed on correct component

sakshi-arora1 (approved):

sakshi-arora1 (approved):

PR #1037, opened 2024-01-12
[FORMS-12594] case handled where file type is not present

deepprakash345:


ui.frontend/src/view/FormFileInputWidget.js
```diff
@@ -414,10 +414,14 @@
                             // check if file names are valid (ie) there are no control characters in file names
                             isInvalidFileName = isCurrentInvalidFileName = true;
                             inValidNamefileNames = currFileName + "," + inValidNamefileNames;
-                        } else if (file.type) {
-                            let isMatch = this.regexMimeTypeList.some(function (rx) {
-                                return rx.test(file.type);
-                            });
+                        } else{
+                            let isMatch=false;
+                            if (file.type) {
```
nit23uec, line 419: this will  lead to error in those scenarios where the customer has explicitly allowed such mimetypes to be uploded but browsers are unable to determine the type. 
browsers typically read the extension to determine the mimetype. So, ideally we should fallback to manual mapping if file.type is undefined. in this case, we should do manual mapping of msft specific mimetypes :
`application/x-msdownload` and a`pplication/vnd.ms-outlook `
to the appropriate extensions.
This will still allow users to upload these files by changing to dummy extensions but the behavior is compatible with that of browsers. Strict checking (by first first few bytes of file to sniff mimetype) could be enforced whenever we introduce server side validation but it is outside of scope of this PR.

nit23uec (changes requested):
inline comment.
and then, lets have a test case also.

ui.frontend/src/view/FormFileInputWidget.js
rismehta: > This will still allow users to upload these files by changing to dummy extensions but the behavior is compatible with that of browsers. Strict checking (by reading first few bytes of file to sniff mimetype) could be enforced whenever we introduce server side validation but it is outside of scope of this PR.

Yes, this is already done on server today for AF v1. Its in backlog for AF v2 based on customer use-case

rismehta (dismissed):
Please add a test case, we already have an existing test suite which you could change.

Also, could you add the failing test under cleanTest / retry, https://github.com/adobe/aem-core-forms-components/blob/master/ui.tests/test-module/specs/replaceaction/replacelayout-container.authoring.spec.js#L176C8-L178C11

ui.frontend/src/view/FormFileInputWidget.js
deepprakash345: added the changes suggested and the testcases

nit23uec (approved):

PR #1057, opened 2024-01-31
Registration of Custom functions x-walk [WIP]

deepprakash345:
## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -301,6 +313,24 @@ public String getLanguageDirection() {
...
+                CUSTOM_FUNCTION_CONFIG_NAME);
+            if (configResource != null) {
+                Resource jcrResource = configResource.getChild(JcrConstants.JCR_CONTENT);
+                if (jcrResource != null) {
+                    ValueMap configValueMap = jcrResource.getValueMap();
+                    String owner = configValueMap.getOrDefault("owner", "").toString();
+                    String repo = configValueMap.getOrDefault("repo", "").toString();
+                    if (!owner.isEmpty() && !repo.isEmpty()) {
+                        String customFunctionUrl = "https://main--" + repo + "--" + owner + ".hlx.live/blocks/form/customFunctions.js";
+                        properties.put(FD_CUSTOM_FUNCTION_URL, customFunctionUrl);
```
nit23uec, line 328: -1 to the use of custom props. pls discuss the approach.

nit23uec (changes requested):
approach needs to be finalised first.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
vdua: We need the implementation on the client. We can bring this URL from the /custom-functions API and then fetch the JS file here.

nit23uec: we should not require the js file, but only the list of custom functions (that will eliminate its parsing on client).
this would make us consistent and also hide the internal details of the url.

PR #1077, opened 2024-02-15
Rich text support

im-shiv:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/TextContent.java
```diff
@@ -0,0 +1,51 @@
...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
+package com.adobe.cq.forms.core.components.models.form;
+
+import org.jetbrains.annotations.Nullable;
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * Interface to represent text as rich content
+ *
+ * @since com.adobe.cq.forms.core.components.models.form 4.8.0
```
rismehta, line 24: Please change the version, we have changed the package-info version

im-shiv: updated

PR #1082, opened 2024-02-22
FORMS-9554 adding dialog changes for rest cloud config

im-shiv:


rismehta (changes requested):
Kindly incorporate test cases for this modification. Additionally, endeavor to employ appropriate semantic CSS selectors to enhance code readability, avoiding the use of generic selectors like div.

rismehta (changes requested):
Kindly investigate the failure in the current test cases; your modifications have caused an issue in an existing test case. This suggests that your change might not be backward compatible.

PR #1096, opened 2024-02-28
fixing the correct response from getDefinition API (visible property …

vdua:
…was coming true always)

- handling visibility in authoring as well

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
@@ -194,9 +194,9 @@ public class AbstractFormComponentImpl extends AbstractComponentImpl implements 
     @JsonInclude(JsonInclude.Include.NON_NULL)
     @Nullable
     public Boolean isVisible() {
-        if (getEditMode()) {
-            return true;
-        }
+        // if (getEditMode()) {
+        // return true;
+        // }
         return visible;
     }
 
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/authoring/css/authoring.css
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/authoring/css/authoring.css
@@ -13,7 +13,3 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  ******************************************************************************/
-
-[data-cmp-visible="false"] {
-    display: initial !important;
-}
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
@@ -14,7 +14,7 @@
  * limitations under the License.
  ******************************************************************************/
 
-html:not(.aem-AuthorLayer-Edit) [data-cmp-visible="false"] {
+.cmp-adaptiveform-container:not(.cq-Editable-dom) > form.cmp-adaptiveform-container [data-cmp-visible="false"] {
     display: none !important;
 }
 .cmp-adaptiveform-container--loading{
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
@@ -14,9 +14,9 @@
  * limitations under the License.
  ******************************************************************************/
 
-[data-cmp-visible="false"] {
+html:not(.aem-AuthorLayer-Edit) [data-cmp-visible="false"] {
     display: none !important;
 }
 .cmp-adaptiveform-container--loading{
-  
+
 }


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -194,9 +194,9 @@ public String getFieldType() {
     @JsonInclude(JsonInclude.Include.NON_NULL)
     @Nullable
     public Boolean isVisible() {
-        if (getEditMode()) {
-            return true;
-        }
+        // if (getEditMode()) {
```
nit23uec, line 197: we should enhance the editMode check to account for EDS url also (`md.html `or whatever)

nit23uec: How are we hitting the aem environment (I assume it's author) with md.html selector?
can you pls share the complete url?

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  ******************************************************************************/
 
-[data-cmp-visible="false"] {
+html:not(.aem-AuthorLayer-Edit) [data-cmp-visible="false"] {
```
nit23uec, line 17: site's css should anyways not be loaded in editor. isn't it the case already?

vdua: This css is loaded as it is part of the runtime.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
vdua: Not in favour of returning a different value based on the request. The client should know whether to honour that value or not.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
nit23uec: we didn;t use to load any runtime lib (except for base I think) in the editor. has that changed?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
nit23uec: this is not true in case of editor load. if there is a public aPI that we can use to recognise the editor load (be it css), then we shoulld use it, other currently, the mechanism in AEM is based on only the request.

vdua: If we look at the [getFormDefinition](https://github.com/adobe/aem-core-forms-components/commit/2893b0ec22e47f2462b468ef50555361bad573fb#diff-20ac9a5b4488d998e45e2359a2fb591a0dfd3561d68842b054e653bffbf1599dR367) code, we can see the following [line](https://github.com/adobe/aem-core-forms-components/commit/2893b0ec22e47f2462b468ef50555361bad573fb#diff-20ac9a5b4488d998e45e2359a2fb591a0dfd3561d68842b054e653bffbf1599dR373)

```
mapper.writerWithView(Views.Publish.class).writeValue(writer, this);
```

We already have a publish view for runtime. Within that Publish view there is another factor which deals with authoring/runtime which is implicit and the developer has no control over it. What that means is if I have to get correct Form Definition in author, there is no API right now. How are workflows getting the correct definition, or they do not require it right now.

We have three different JSON right now
1. author view
2. publish view for authoring
3. publish view for publishing

Ideally we should have only author view / publish view.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  ******************************************************************************/
 
-[data-cmp-visible="false"] {
+.cmp-adaptiveform-container:not(.cq-Editable-dom) > form.cmp-adaptiveform-container [data-cmp-visible="false"] {
```
vdua, line 17: is a wrapper div added by container always. I think there was a mechanism to get rid of that wrapper div (in which case this will fail).

rismehta: Why is runtime client library referring editable CSS selector ?

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/authoring/css/authoring.css
```diff
@@ -13,7 +13,3 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  ******************************************************************************/
-
-[data-cmp-visible="false"] {
```
rismehta, line 17: This modification is backward in-comatible. Ideally, we should address the root issue in the original code instead of fixing the CSS. Removing this data attribute could impact custom core components that rely on it, and not doing so may result in their hidden components remaining hidden rather than being visible during authoring.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
rismehta: I believe it's unlikely for us to process the JSON from the publish view at runtime because it relies on fix from WCM core component. It's important to mention that we haven't conducted test cases for the runtime with the publish view.

rismehta (changes requested):
Upon reviewing the code once more, I found that the `isEditMode` function returns true only for specific use-cases, and this was intentionally considered during the initial implementation. For your particular scenario, it should ideally return false. The reason for it returning true in your case is still unclear. Can you elaborate on this ?

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
rismehta: We don't load runtime lib in editor.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
nit23uec: One possible solution for java API working: We can fix it having a custom seralizer for this property. We have to use  SerializerProvider#getActiveView while serializing. If the view is Author or Publish, we return the actual property as it is. If there is no view, then the default fallback mechanism based on request applies.

nit23uec:
There are only two ways on a high level to handle Editor vs Runtime cases:
Either deliver different json and let runtime code be transparent of this handling for both editor/runtime modes or deliver same json and handle it in runtime code. By runtime code, I'm referring to the client side code that consumes the json and renders the form. We chose the former approach in core components as it makes runtime code consistent and we don't have to worry about editorMode handling at multiple places (including rule editor events) in the runtime code. Imagine visible property returning false and then client has written code to subscribe on rule editor events that does not makes use of data-cmp-visible attribute.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/site/css/formcontainerview.css
nit23uec: On further analysis: we actually load runtime css in the editor to have the look and feel same for WYSIWYG authoring.

rismehta:
This is already fixed. Closing the PR

PR #1102, opened 2024-03-01
Fixed updateRequired for checkboxgroup and radiobutton

ravise5:
## Types of changes

## Checklist:

### DIFF
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/checkboxgroup/v1/checkboxgroup/clientlibs/site/js/checkboxgroupview.js
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/checkboxgroup/v1/checkboxgroup/clientlibs/site/js/checkboxgroupview.js
@@ -179,6 +179,12 @@
                 }
             });
         }
+        updateRequired(required, state) {
+            if (this.widget) {
+                this.toggle(required, "required");
+                this.element.setAttribute("data-cmp-required", required);
+            }
+        }
     }
 
 
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/checkboxgroup/v1/checkboxgroup/clientlibs/site/js/checkboxgroupview.js
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/checkboxgroup/v1/checkboxgroup/clientlibs/site/js/checkboxgroupview.js
@@ -181,7 +181,7 @@
         }
         updateRequired(required, state) {
             if (this.widget) {
-                this.toggle(required, "required");
+                this.element.toggleAttribute("required", required);
                 this.element.setAttribute("data-cmp-required", required);
             }
         }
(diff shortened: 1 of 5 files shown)


sakshi-arora1 (approved):

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -396,10 +396,12 @@ class FormFieldBase extends FormField {
         if (this.widget) {
             this.toggle(required, "required");
             this.element.setAttribute(Constants.DATA_ATTRIBUTE_REQUIRED, required);
-            if (required === true) {
-                this.widget.setAttribute("required", "required");
-            } else {
-                this.widget.removeAttribute("required");
+            if(!(this.widget instanceof NodeList)) { // filtering radiobutton and checkboxGroup
```
nit23uec, line 399: NodeList is a generic interface that applies to more than radiobutton or checkbox elements. Any component specific handling should be done in respective component views.

nit23uec (changes requested):
Pls check comments

ui.frontend/src/view/FormFieldBase.js
ravise5: Reason I kept it in top level was to reduce the code duplication for checkbox and radiobuttons. 

And when a component has multiple widgets which one do we set as required?  I believe we set it in parent div only so isn't this the generic behaviour?

nit23uec: Think about a (custom) component say telephone number that capture country code and the number in a string and numeric/tel fields respectively. In this case both of the nodes in the parent div may be required to be marked required. FormFieldBase should offer any minimal things as it can't be changed by  customization.

rismehta: @ravise5 Can you move this logic specific to radio button and check box view layer to keep the base layer simple ?

ravise5: Okay.

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -396,11 +396,11 @@ class FormFieldBase extends FormField {
         if (this.widget) {
             this.toggle(required, "required");
             this.element.setAttribute(Constants.DATA_ATTRIBUTE_REQUIRED, required);
-            if (required === true) {
-                this.widget.setAttribute("required", "required");
-            } else {
-                this.widget.removeAttribute("required");
-            }
+                if (required === true) {
```
nit23uec, line 399: minor: indentation needs to be fixed.

nit23uec: as an improvement, it'll be good to use `toggleAttribute` method in base.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/checkboxgroup/v1/checkboxgroup/clientlibs/site/js/checkboxgroupview.js
```diff
@@ -179,6 +179,12 @@
                 }
             });
         }
+        updateRequired(required, state) {
+            if (this.widget) {
+                this.toggle(required, "required");
+                this.element.setAttribute("data-cmp-required", required);
```
nit23uec, line 185: `this.element.toggleAttribute`

rismehta (approved):

PR #1107, opened 2024-03-04
[FORMS-13460] Fixing visible, readonly, required and enabled property

rismehta:
- JSON would always return the same state irrespective of edit mode
- all data attributes of these properties were incorrect when page source is inspected, fixed the same
- we still load runtime client libraries on edit mode due to backward compatibility @review @vdua
DOD(Yes)

## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFieldImpl.java
```diff
@@ -124,16 +126,24 @@ public abstract class AbstractFieldImpl extends AbstractBaseImpl implements Fiel
     private Resource resource;
 
     @Override
-    @JsonInclude(JsonInclude.Include.NON_NULL)
     @Nullable
     public Boolean isReadOnly() {
+        return readOnly != null ? readOnly : Boolean.FALSE;
+    }
+
+    @JsonProperty("readOnly")
```
rismehta, line 134: This is added to keep the JSON payload minimal, we only add readOnly attribute if its present in JCR else we don't return anything

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -191,15 +193,17 @@ public String getFieldType() {
      * @since com.adobe.cq.forms.core.components.models.form 0.0.1
      */
     @Override
-    @JsonInclude(JsonInclude.Include.NON_NULL)
     @Nullable
     public Boolean isVisible() {
-        if (getEditMode()) {
-            return true;
-        }
+        return visible == null || visible;
```
rismehta, line 198: data-cmp-visible when inspected in page source was having its value as false due to this, https://github.com/adobe/aem-core-forms-components/blob/master/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/textinput/v1/textinput/textinput.html#L28. I have fixed other such issues related to this

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/README.md
```diff
@@ -63,6 +63,15 @@ BLOCK cmp-adaptiveform-container
     ELEMENT cmp-adaptiveform-container__wrapper
 ```
 
+In edit mode, the BEM structure includes a modifier for the cmp-adaptiveform-container block,
```
rismehta, line 66: In Edit mode, we utilize a distinct BEM modifier to indicate the editing state. To ensure simplicity for clients, it is crucial to maintain a consistent JSON payload. While properties can be added or removed based on the JSON view, the values of a property should remain consistent across various tiers, else client's (like EDS) would have to contribute to our code base by whitelisting things which is not scalable

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
nit23uec: can you pls test this change by marking an accordion as hidden?

nit23uec: There's one issue with the css approach:
It's upto the consumer of the json to consume the `visible` property. the customer need not use the `data-cmp-visible` in her customization and can write a simple code like:
```
if (visible === false) {
 this.element.style.display = "none";
}
```
She can even use a different data attribute to set for visible property in her custom code and she might rely on a css for the same.
Now she will have to keep in mind to not include this js/css file in the authoring otherwise, it will be an issue that's hard to debug for her.

rismehta: We've implemented a CSS rule that utilizes the !important declaration, indicating its priority over other rules. Here's the rule:

`
form:not(.cmp-adaptiveform-container--edit) [data-cmp-visible="false"] {
    display: none !important;
}
`

If someone opts not to use the `data-cmp-visible` attribute, the visibility behavior may not function correctly. This is because, at the view layer, we rely on these data attributes to deduce the component's visibility. Currently, the attribute is optional, and the form still operates, but visibility won't function as expected. We've made only a few attributes mandatory, governing the model's initialization.

rismehta:
Merging since this is blocking EDS use-case, if any review comments are pending would take it up later.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
vdua: Any reason for customising by setting `style.display` instead of `data-cmp-visible` that is documented. I don't see a strong reason, but if they are then they can handle the authoring case as well.

PR #1128, opened 2024-03-15
old enum names string array support added

im-shiv:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/OptionsConstraint.java
```diff
@@ -69,6 +69,16 @@ default boolean isEnforceEnum() {
      * @return the list of enum names
      * @since com.adobe.cq.forms.core.components.models.form 5.4.0
      */
-    @JsonProperty("enumNames")
+    @JsonIgnore
     TextContent[] getEnumNamesAsTextContent();
```
rismehta, line 73: Can you remove this function itself, I don't think anyone would be using it yet. Let's have one public facing API for this.

rismehta (changes requested):
One comment, please check

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/OptionsConstraint.java
im-shiv: we are using it in sightly to enable rich text, eg: https://github.com/adobe/aem-core-forms-components/blob/master/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/checkboxgroup/v1/checkboxgroup/widget.html#L28

rismehta: Please use the new API, it does not make sense to have two API's for the same thing.

im-shiv: got it, removed old APi

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/OptionsConstraint.java
```diff
@@ -63,12 +63,12 @@ default boolean isEnforceEnum() {
...
+     * Returns an array of String or TextContent to be displayed to the end user based on the richText flag.
      * The length of enum and enumNames array must match
      *
      * @return the list of enum names
-     * @since com.adobe.cq.forms.core.components.models.form 5.4.0
+     * @since com.adobe.cq.forms.core.components.models.form 5.5.0
      */
     @JsonProperty("enumNames")
-    TextContent[] getEnumNamesAsTextContent();
+    Object[] getEnumNamesArray();
```
rismehta, line 73: Let's keep this array of string, remove the `TextContent` completely from the java code. UI could still have the same flow, let's change the spec too and fix this.

rismehta (changes requested):
check comments

PR #1148, opened 2024-03-27
Enabling custom extension in view layer

rismehta:
## Types of changes

## Checklist:


nit23uec:
-1 for exposing individual view layer classes to global names pace. They have been deliberately kept as internal to reduce maintenance and chances of breaking brekaward compatibility.
The layer is already thin and the developer can just copy paste the code in the cuts componem if required.
What's the exact issue you're facing to customize the component?
Also: pls avoid merging PRs without even a single review.

rismehta:
>-1 for exposing individual view layer classes to global names pace. They have been deliberately kept as internal to reduce maintenance and chances of breaking backward compatibility. The layer is already thin and the developer can just copy paste the code in the custom component if required.

Customers have a use case to extend individual components, hence their exposure. While developers can resort to copying and pasting, any alterations in the existing view layer, such as fixing minor bugs, would necessitate the repetition of the copy-and-paste process after each upgrade. This is not a seamless developer experience.

> Also: pls avoid merging PRs without even a single review.

Given the magnitude of this alteration, it's worth noting that the changes have already undergone offline review by @vdua .why a significant change like this would be merged without review.

> What's the exact issue you're facing to customize the component?

Have you had a chance to review the list of changes to see if the customer's use case is included in the test collateral?

sakshi-arora1 (commented):
Few comments:
1. This increases maintenance for us. This will cause an overhead. This also may result in breaking of the custom components with changes to the view layer.
2. Namespace should be FormView.Accordion.v1 rather than FormView.v1.Accordion
3. Tests for custom component in repeatable panel is missing.

rismehta:
>-1 for exposing individual view layer classes to global names pace.

@nit23uec Conversations held offline with @vdua and @sakshi-arora1 have led to the decision not to expose the namespace at this time. Instead, we will allow customers to copy and paste the code. In the event of any changes during upgrades, developers will need to manually copy and paste the new code. 

@barshat7 Please remove the code which exposes the individual component view to global namespace.

PR #1169, opened 2024-04-08
FORMS-13851 Support for language in date picker and number input

rismehta:
@review @vdua
DOD(Yes)

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
@@ -85,6 +85,7 @@ public class FormMetaDataDataSourceServlet extends AbstractDataSourceServlet {
     public enum FormMetaDataType {
         SUBMIT_ACTION("submitAction"),
         PREFILL_ACTION("prefillServiceProvider"),
+        LANG("lang"),
         FORMATTERS("formatters");
 
         private String value;
@@ -155,6 +156,7 @@ public class FormMetaDataDataSourceServlet extends AbstractDataSourceServlet {
             Iterator<FormsManager.ComponentDescription> metaDataList = null;
             switch (type) {
                 case FORMATTERS:
+                case LANG:
                     ContentPolicy policy = ComponentUtils.getPolicy((String) request.getAttribute(Value.CONTENTPATH_ATTRIBUTE),
                         resourceResolver);
                     resources.add(getResourceForDropdownDisplay(resourceResolver, "Select", ""));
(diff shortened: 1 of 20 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
```diff
@@ -155,22 +156,26 @@ private List<Resource> getDataSourceResources(HttpServletRequest request, Resour
...
                     ContentPolicy policy = ComponentUtils.getPolicy((String) request.getAttribute(Value.CONTENTPATH_ATTRIBUTE),
                         resourceResolver);
                     resources.add(getResourceForDropdownDisplay(resourceResolver, "Select", ""));
                     if (policy != null) {
                         ValueMap props = policy.getProperties();
                         if (props != null) {
                             for (Map.Entry<String, Object> entry : props.entrySet()) {
-                                if (entry.getKey().startsWith(ALLOWED_FORMAT)) {
+                                if ((type.equals(FormMetaDataType.FORMATTERS) && entry.getKey().startsWith(ALLOWED_FORMAT)) ||
+                                    (type.equals(FormMetaDataType.LANG) && entry.getKey().startsWith(FormMetaDataType.LANG.getValue()))) {
```
vdua, line 168: I would suggest making private functions and using that here for readability, something like below
```
if (isFormattersPolicy(type, entry) || isLangPolicy(type, entry))
```

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
```diff
@@ -155,6 +158,11 @@ public boolean isTooltipVisible() {
         return tooltipVisible;
     }
 
+    @Override
+    public String getLang() {
+        return lang;
```
vdua, line 163: we are returning lang for every field then.

bundles/af-core/src/test/resources/form/datepicker/test-content.json
```diff
@@ -17,6 +17,8 @@
     "enabled": false,
     "readOnly": true,
     "required": false,
+    "lang" : "it-IT",
+    "langDisplayValue" : "it-IT",
```
vdua, line 21: what is langDisplayValue ?

ui.tests/test-module/specs/numberinput/numberinput.runtime.spec.js
```diff
@@ -248,3 +248,33 @@ describe("Form with Number Input", () => {
...
+
+    it("display pattern on numeric input should update the display value based on field language", () => {
+        const [numberInput6, numberInput6FieldView] = Object.entries(formContainer._fields)[5];
+        const input = "121212";
+        let model = numberInput6FieldView.getModel();
+
+        cy.get(`#${numberInput6}`).find("input").clear().type(input).blur().then(x => {
+            expect(Number(model.getState().value)).to.equal(Number(input));
+// Assert that the input value contains "121" and "212,000" regardless of the space type
+                cy.get(`#${numberInput6}`).find('input').invoke('val').should('match', /121.*212,000/);
```
vdua, line 277: value should be `121.121,00` why are we using regex to match it ?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
rismehta: yes, but only if it is set, it would be returned.

ui.tests/test-module/specs/numberinput/numberinput.runtime.spec.js
rismehta: There is some space in the display pattern, hence a regex is required to match

vdua: Ideally space should not come. With space the the correct regex should be `121\.\s*121,000`

vdua (approved):

PR #1171, opened 2024-04-08
datepicker readonly does not work when edit or display format set

barshat7:
## Types of changes

## Checklist:


ui.tests/test-module/specs/datepicker/datepicker.runtime.spec.js
```diff
@@ -268,4 +268,18 @@ describe("Form Runtime with Date Picker", () => {
             });
         });
     });
+
+    it.only("should not show calendar widget if marked readonly", () => {
```
rismehta, line 272: Please remove only

rismehta (changes requested):
minor comment, else changes looks good now. Can you also add test case to update read only via JS API in the cypress test case, so that the other piece of code also has the required coverage ?

rismehta (approved):

PR #1180, opened 2024-04-10
FORMS-13760 Date picker display value not working during prefill

rismehta:
@review @vdua
DOD(Yes)

## Types of changes

## Checklist:

### DIFF
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/datepicker/v1/datepicker/clientlibs/site/js/datepickerview.js
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/datepicker/v1/datepicker/clientlibs/site/js/datepickerview.js
@@ -77,15 +77,15 @@
                 if (this.widgetObject == null) {
                     this.widgetObject = new DatePickerWidget(this, this.getWidget(), model);
                 }
-                if (this.widgetObject.getValue() !== '') {
-                    this._model.value = this.widgetObject.getValue();
+                if (this.isActive()) {
+                    this.widgetObject.setValue(model.value);
+                } else {
+                    this.widgetObject.setDisplayValue(model.value);
                 }
                 this.widgetObject.addEventListener('blur', (e) => {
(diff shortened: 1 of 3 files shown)


barshat7 (approved):

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/datepicker/v1/datepicker/clientlibs/site/js/datepickerview.js
```diff
@@ -77,15 +77,15 @@
                 if (this.widgetObject == null) {
                     this.widgetObject = new DatePickerWidget(this, this.getWidget(), model);
                 }
-                if (this.widgetObject.getValue() !== '') {
-                    this._model.value = this.widgetObject.getValue();
+                if (this.isActive()) {
+                    this.widgetObject.setValue(model.value);
```
sakshi-arora1, line 81: Why are we setting widget value in setting of model? Ideally this should be the job of updateValue not setModel
Earlier model value was being set via widget, you have al together removed that code..

rismehta: > Earlier model value was being set via widget

It was redundant because the model value is already established upon initializing the datePickerWidget. If you observe the third argument, it contains the model.

> Why are we setting widget value in setting of model?

Ideally, we should utilize updateValue. However, during prefill/initialize, updateValue lacks initialization of the widgetObject, resulting in failure. Although I considered making this change of `initializing widgetObject in updateValue like we do in number input` but it would require significant refactoring, which isn't necessary at this juncture.

sakshi-arora1: Can you open a JIRA to correct this flow, to align with other views

PR #1198, opened 2024-04-22
adding source property in sling model of image component to get dam asset path in headless json

deepprakash345:


bundles/af-core/src/test/resources/form/image/exporter-image-parsedSrc.json
```diff
@@ -0,0 +1,18 @@
+{
+  "id": "image-7cfd7f1fe4",
+  "fieldType": "image",
+  "name": "abcd",
+  "visible": false,
+  "altText": "abc",
+  "value": "/content/image-customized.img.png",
+  "source": "/content/image-customized.img.png",
```
rismehta, line 8: The schema validation test case succeeded because it currently only verifies required properties for specific field types. However, ideally, we should also review and possibly update the specification for source/value property name. Can you start a conversation on our internal Slack channel to finalize the property name? In my opinion, the source should be a custom property that points to the DAM image, given that this detail is specific to AEM. The specification should mandate only a single property for an image, which is its URL (whether the property name should be value or source, we need to decide the same, but only one property should be enforce by spec)

rismehta (changes requested):
check comments

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/StaticImage.java
```diff
@@ -49,6 +49,16 @@ default String getValue() {
         return null;
     }
 
+    /**
+     * Retrieves the parsed img value to be displayed.
+     *
+     * @return the parsed img value to be displayed, or {@code null} if no value can be returned
+     * @since com.adobe.cq.forms.core.components.models.form 0.0.1;
+     */
+    default String getSource() {
```
rismehta, line 58: Why do you need this API now, can you remove it

rismehta (changes requested):
check comments

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/StaticImage.java
deepprakash345: removed

rismehta (approved):

PR #1210, opened 2024-04-30
FORMS-14018 - Fix for dropdown not working when enumNames are not provided

TalmizAhmed:
Currently, when not providing enumNames in the dialog, the values from enums are copied to it in js before being submitted to jcr. An assumption of the constant presence of enumNames was made in the sling model code. But we have a use case in UE where enums can be provided without enumNames, which breaks this, hence the fix.

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v1/form/DropDownImplTest.java
+++ b/bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v1/form/DropDownImplTest.java
@@ -61,6 +61,9 @@ public class DropDownImplTest {
 
     private static final String PATH_DROPDOWN_WITH_DUPLICATE_ENUMS = CONTENT_ROOT + "/dropdown-duplicate-enum";
     private static final String PATH_DROPDOWN_FOR_INSERTION_ORDER = CONTENT_ROOT + "/dropdown-insertion-order";
+    private static final String PATH_DROPDOWN_FOR_NO_ENUM_NAMES = CONTENT_ROOT + "/dropdown-without-enumNames";
+    private static final String PATH_DROPDOWN_FOR_LESSER_ENUM_NAMES = CONTENT_ROOT + "/dropdown-with-lesser-enumNames";
+    private static final String PATH_DROPDOWN_FOR_EXTRA_ENUM_NAMES = CONTENT_ROOT + "/dropdown-with-extra-enumNames";
 
     private final AemContext context = FormsCoreComponentTestContext.newAemContext();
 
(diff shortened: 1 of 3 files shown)


rismehta:
Please don't add corp JIRA in github, only refer the issue link

bundles/af-core/src/test/resources/form/dropdown/test-content.json
```diff
@@ -211,5 +211,75 @@
...
+      "5",
+      "6",
+      "7",
+      "8",
+      "9"
+    ],
+    ":type": "forms-components-examples/components/form/dropdown"
+  },
+  "dropdown-with-extra-enumNames": {
+    "id": "dropdown-f0a2750c3b",
```
nit23uec, line 255: Pls add the test case where there are no enumNames but only enum

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v1/form/DropDownImplTest.java
```diff
@@ -61,6 +61,9 @@ public class DropDownImplTest {
 
     private static final String PATH_DROPDOWN_WITH_DUPLICATE_ENUMS = CONTENT_ROOT + "/dropdown-duplicate-enum";
     private static final String PATH_DROPDOWN_FOR_INSERTION_ORDER = CONTENT_ROOT + "/dropdown-insertion-order";
+    private static final String PATH_DROPDOWN_FOR_NO_ENUM_NAMES = CONTENT_ROOT + "/dropdown-without-enumNames";
```
nit23uec, line 64: I could not see any test content with no enum names. Is it a miss?

TalmizAhmed: @nit23uec This was intentional, I had a test case for this [(see here)](https://github.com/adobe/aem-core-forms-components/blob/a302c74e46954b719a66578179e6c8b7da3cce22/bundles/af-core/src/test/resources/form/dropdown/test-content.json#L215) but had to remove it since supporting it would mean adding non backward compatible changes.
There are already test cases that are present for other components (checkbox, TnC) that require enumNames to remain empty even when enums are given. [(Ref)](https://github.com/adobe/aem-core-forms-components/blob/395f3069fbf9d4b6541a2f624e27dacf31b9c79b/bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v1/form/CheckBoxImplTest.java#L313)

The only solution left is to unblock enums from being returned when enumNames is empty which is what the current code does.

bundles/af-core/src/test/resources/form/dropdown/test-content.json
TalmizAhmed: Answered here: https://github.com/adobe/aem-core-forms-components/pull/1210#discussion_r1587050975

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v1/form/DropDownImplTest.java
TalmizAhmed: @nit23uec I had [a test case](https://github.com/adobe/aem-core-forms-components/blob/a302c74e46954b719a66578179e6c8b7da3cce22/bundles/af-core/src/test/resources/form/dropdown/test-content.json#L215) but removed it intermittently, since I saw there could be backward incompatible changes. Came up with an alternative fix that evades that route. Added the test case back.

rismehta (approved):

PR #1237, opened 2024-05-16
Supporting getFormDefinition for UE

TalmizAhmed:
## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
@@ -15,6 +15,9 @@
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
 package com.adobe.cq.forms.core.components.internal.models.v2.form;
 
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.Writer;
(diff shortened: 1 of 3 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -362,4 +367,18 @@ public String getName() {
         return FormContainer.super.getName();
     }
 
+    @JsonIgnore
```
nit23uec, line 370: jsonignore should not be required in impl, I guess.

TalmizAhmed: I see it being used in other methods above of the same class, and so maybe it is?
cc: @rismehta

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormContainer.java
```diff
@@ -362,4 +362,14 @@ default void visit(Consumer<ComponentExporter> callback) throws Exception {}
...
         return null;
     }
+
+    /**
+     * Returns the form container's json.
+     * 
+     * @return A String that consists of the form container json
+     */
+    @JsonIgnore
+    default String getFormDefinition() {
```
rismehta, line 372: This is already part of FormStructureParser and it is merged. Why do we have another PR for this ?

rismehta (changes requested):
incorrect PR

PR #1264, opened 2024-06-10
UE: Custom property support for Universal Editor

TalmizAhmed:
## Description
Any property present in the jcr node that is not a part of the ReservedProperties list will become a custom key value pair within the properties object of the component's json.

## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -483,14 +486,32 @@ public boolean isEmpty(SerializerProvider provider, String value) {
      * @return {@code Map<String, String>} returns all custom property key value pairs associated with the resource
      */
     private Map<String, String> getCustomProperties() {
+        Map<String, String> customProperties = new HashMap<>();
+        Map<String, String> templateBasedCustomProperties;
```
rismehta, line 490: Property type can be other than string too, for example, existing customer props could be boolean, number (int / decimal) or string. We need to handle all with these changes.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -483,14 +486,32 @@ public boolean isEmpty(SerializerProvider provider, String value) {
...
+        List<String> excludedPrefixes = Arrays.asList("fd:", "jcr:", "sling:");
```
rismehta, line 491: Please add test cases in exporter for these prefixes. What if there is a property which starts with `fd:` and I want it to be a custom property. I guess the below code won't add this in custom properties.

Will individual classes override it ?

rismehta (changes requested):
check comments

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
TalmizAhmed: We had a discussion in the spec review that we wont be supporting types other than string to keep things aligned with the same reasons mentioned for template based custom properties.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
TalmizAhmed: By default any property with the `fd:` prefix will have to be handled separately by each class. Just like we have done for `fd:viewType` [here](https://github.com/adobe/aem-core-forms-components/blob/4f37812634959bb0d4e7651add6a47ff287328d3/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java#L127)

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
rismehta: There can be existing customer properties with other types

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
```diff
@@ -0,0 +1,171 @@
...
+    public static final String PN_REDIRECT = "redirect";
+    public static final String PN_PREFILL_SERVICE = "prefillService";
+    public static final String PN_SPEC_VERSION = "specVersion";
+    public static final String PN_RICH_TEXT = "richText";
+
+    public static List<String> getNodeProperties() {
+        List<String> nodeProperties = new ArrayList<>();
+        Field[] fields = ReservedProperties.class.getDeclaredFields();
+
+        for (Field field : fields) {
```
barshat7, line 159: Can we store the properties in a singleton class so that the below loop with reflection doesn't get executed every single time

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -483,14 +486,32 @@ public boolean isEmpty(SerializerProvider provider, String value) {
...
+
+        List<String> reservedProperties = ReservedProperties.getNodeProperties();
```
barshat7, line 493: Can we convert this to a hashset? As we are using the .contains inside another loop.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
nit23uec: there's no technical limitation but increased complexity in adding `types` support. The discussion we had in review a year back was we shouldn't map the custom prop types directly to jcr types and decoupling this would further require a type property in custom property map that would complicate things from usability perspective too.
Until a requirement comes, we should support string props, IMO.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
nit23uec: +1. lets have a static field here to store the props.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
nit23uec: or, for common props, we could handle it just like it's handled for fd:dor. (after `getCustomProperties` method is invoked in `getProperties`)

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
nit23uec: +1

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
rismehta: Even today, we map jcrType to the CRISPR JSON type. According to the CRISPR specification under the AEM appendix, custom properties can have types other than strings. We already handle these cases effectively.

> would further require a type property in custom property map that would complicate things from usability perspective too.

I don't think you require a type property, I am just talking about this simple change, which will not break even existing customers,

```
  Map<String, Object> nodeBasedCustomProperties = resourceMap.entrySet()
            .stream()
            .filter(entry -> isAllowedType(entry.getValue())
                && !reservedProperties.contains(entry.getKey())
                && excludedPrefixes.stream().noneMatch(prefix -> entry.getKey().startsWith(prefix)))
            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```

nit23uec: I think it's fine to have the above code. @TalmizAhmed can you pls verify it for all primitive types supported by UE?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
rismehta: > or, for common props, we could handle it just like it's handled for fd:dor. (after `getCustomProperties` method is invoked in `getProperties`)

Yes, if the property is common for all field, it could be done this way too.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
rismehta, line 493: @TalmizAhmed Why did your initial CL even have this 😄  This un-necessary computation could have being saved. Good catch @barshat7 !

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -483,14 +486,32 @@ public boolean isEmpty(SerializerProvider provider, String value) {
...
         try {
-            return Optional.ofNullable(this.resource.adaptTo(CustomPropertyInfo.class))
+            templateBasedCustomProperties = Optional.ofNullable(this.resource.adaptTo(CustomPropertyInfo.class))
                 .map(CustomPropertyInfo::getProperties)
                 .orElse(Collections.emptyMap());
         } catch (NoClassDefFoundError e) {
             logger.info("CustomPropertyInfo class not found. This feature is available in the latest Forms addon. Returning an empty map.");
             return Collections.emptyMap();
         }
+        if (!templateBasedCustomProperties.isEmpty()) {
```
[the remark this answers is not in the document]
TalmizAhmed: Removing this check since there is already a case for null and empty.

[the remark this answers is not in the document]
TalmizAhmed: Test cases for template based custom properties are present in cq-guides and as an e2e cypress test here that would cover this.
As for node based, I have added custom properties in the test content of the current test cases.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
TalmizAhmed: Will support String, Boolean, Calendar, Long, BigDecimal and their respective arrays.

bundles/af-core/src/test/resources/form/textinput/test-content.json
```diff
@@ -4,7 +4,17 @@
...
+    "customPropString": "test-custom-string",
+    "customPropStringArray": ["a", "b"],
+    "customPropBool": true,
+    "customPropBoolArray": true,
+    "customPropLong": 45,
+    "customPropLongArray": [345, 576],
+    "customPropDecimal": 45.67,
+    "customPropDecimalArray": [45.67, 90.34],
+    "customPropDate": "2022-09-13T09:00:00.000Z",
+    "customPropDateArray": ["2022-09-13T09:00:00.000Z", "2024-06-17T02:30:00.000+05:30"]
```
TalmizAhmed, line 17: @rismehta Here are the test cases for all primitive types.

bundles/af-core/src/test/resources/form/textinput/test-content.json
```diff
@@ -4,7 +4,20 @@
...
+    "customPropBoolArray": true,
+    "customPropLong": 45,
+    "customPropLongArray": [345, 576],
+    "customPropDecimal": 45.67,
+    "customPropDecimalArray": [45.67, 90.34],
+    "customPropDate": "2022-09-13T09:00:00.000Z",
+    "customPropDateArray": ["2022-09-13T09:00:00.000Z", "2024-06-17T02:30:00.000+05:30"],
+    "fd:accidentalPrefix": "shouldNotBeInProperties",
+    "sling:accidentalPrefix": "shouldNotBeInProperties",
+    "jcr:accidentalPrefix": "shouldNotBeInProperties"
```
TalmizAhmed, line 20: @rismehta Here are the test cases for all primitive types and for excluded prefixes.
cc: @nit23uec

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
TalmizAhmed: Added `testGetCustomProperties` in TextInputImplTest

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -16,14 +16,8 @@
...
-import java.util.AbstractMap;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.math.BigDecimal;
+import java.util.*;
```
nit23uec, line 20: pls avoid importing wildcards.

bundles/af-core/src/test/resources/form/recaptcha/exporter-recaptcha.json
```diff
@@ -31,7 +31,9 @@
         "type": "image",
         "version": null
       }
-    }
+    },
+    "mandatory": "true",
```
nit23uec, line 35: mandatory is not part of schema. how is that is exporter?

nit23uec, line 35: mandatory translates to required as first class prop. why is this part of custom prop?

TalmizAhmed: Did not see an instance of it set in the code way back. Later found out we have a lot of these and others via content.xml, of which I forgot to put this one back.
Adding back to reserve list now.

nit23uec (approved):

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -477,20 +482,43 @@ public boolean isEmpty(SerializerProvider provider, String value) {
         }
     }
 
+    private boolean isAllowedType(Object value) {
+        return value instanceof String || value instanceof String[] || value instanceof Boolean || value instanceof Boolean[]
```
rismehta, line 486: This can be simplified using streams,

```
  return Stream.of(
            String.class, String[].class,
            Boolean.class, Boolean[].class,
            Calendar.class, Calendar[].class,
            BigDecimal.class, BigDecimal[].class,
            Long.class, Long[].class
    ).anyMatch(type -> type.isInstance(value));
    ```

rismehta (approved):

PR #1267, opened 2024-06-11
UE: Encoding HTML in Tooltip and Description using XSSAPI

TalmizAhmed:
## Description
UE allows rich text to be added in these fields, and since they will contain HTML tags, the form json embedded in the code tags is split by the tags present inside these fields and causes the eds rendition of the form to break when they are not encoded properly

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
@@ -19,6 +19,10 @@ import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
 
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.models.annotations.Model;
@@ -31,7 +35,6 @@ import org.slf4j.LoggerFactory;
 
 import com.adobe.cq.forms.core.components.models.form.FormContainer;
 import com.adobe.cq.forms.core.components.models.form.FormStructureParser;
-import com.adobe.cq.forms.core.components.util.AbstractBaseImpl;
 import com.adobe.cq.forms.core.components.util.ComponentUtils;
 import com.adobe.cq.forms.core.components.views.Views;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -120,7 +123,7 @@ public class FormStructureParserImpl implements FormStructureParser {
         FormContainer formContainer = resource.adaptTo(FormContainer.class);
         try {
             ObjectMapper mapper = new ObjectMapper();
-            mapper.registerModule(new SimpleModule().addSerializer(String.class, new AbstractBaseImpl.EncodeHTMLSerializer()));
+            mapper.registerModule(new SimpleModule().addSerializer(String.class, new FormStructureParserImpl.EncodeHTMLSerializer()));
             Writer writer = new StringWriter();
             // return publish view specific properties only for runtime
             mapper.writerWithView(Views.Publish.class).writeValue(writer, formContainer);
@@ -129,5 +132,14 @@ public class FormStructureParserImpl implements FormStructureParser {
             logger.error("Unable to generate json from resource");
         }
         return result;
+    }
+    private static class EncodeHTMLSerializer extends JsonSerializer<String> {
+        @Override
+        public void serialize(String value, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
+            if (value != null) {
+                String escapedValue = StringEscapeUtils.escapeHtml4(value);
+                jsonGenerator.writeString(escapedValue);
+            }
+        }
     }
 }
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
@@ -31,9 +31,11 @@ import org.slf4j.LoggerFactory;
 
 import com.adobe.cq.forms.core.components.models.form.FormContainer;
 import com.adobe.cq.forms.core.components.models.form.FormStructureParser;
+import com.adobe.cq.forms.core.components.util.AbstractBaseImpl;
 import com.adobe.cq.forms.core.components.util.ComponentUtils;
 import com.adobe.cq.forms.core.components.views.Views;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
 
 @Model(
     adaptables = { SlingHttpServletRequest.class, Resource.class },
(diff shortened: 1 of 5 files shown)


TalmizAhmed:
I am aware that this is a backward incompatible change as the value of the tooltip and description is changing in the JSON. What version would I have to bump?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
```diff
@@ -156,6 +164,10 @@ public Label getLabel() {
 
     @Override
     public @Nullable String getTooltip() {
+        if (xssApi != null) {
+            logger.info("[AF] XSS API has been used to encode tooltip HTML.");
+            return translate(PN_TOOLTIP, xssApi.encodeForHTML(tooltip));
```
nit23uec, line 169: Although we are saving a rich text through dialog for these props, there seems no existing test case for rich text.
what's the current status for these 2 props? is translation working for rich text?

rismehta: Translation should work. @TalmizAhmed Please enhance the translation collateral /spec which exist today to incorporate this use-case

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
```diff
@@ -156,6 +164,10 @@ public Label getLabel() {
 
     @Override
     public @Nullable String getTooltip() {
+        if (xssApi != null) {
+            logger.info("[AF] XSS API has been used to encode tooltip HTML.");
```
rismehta, line 168: You need to handle the else case too, when xssAPI Is not available. Please get ride of info logs from here

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
```diff
@@ -228,6 +240,10 @@ public String getHtmlScreenReaderText() {
     @Override
     @Nullable
     public String getDescription() {
+        if (xssApi != null) {
+            logger.info("[AF] XSS API has been used to encode description HTML.");
+            return translate(PN_DESCRIPTION, xssApi.encodeForHTML(description));
```
rismehta, line 245: encodeForHTML is necessary only for Sightly, so it is best to use the xssAPI directly within the Sightly file.

For JSON serialization, the Jackson library automatically handles encoding when converting Java objects to JSON.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
rismehta: in v2, translation works via key and not string match, hence even with encoding it would work. But fundamentally the PR is backward incompatible, its needs to changed

TalmizAhmed: @rismehta Translation does not work with rich text currently
Screenshot 2024-06-11 at 2 47 12 PM

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
TalmizAhmed: It unfortunately doesn't. The tooltip and description come as the following in json which is causing the break:
```
{
...
tooltip: "<p>Hello</p>",
...
}
```

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
TalmizAhmed: Else cases are already handled right below the if block since there's a return inside.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
rismehta: It works, we already have a test case. You might have missed some steps. Please sync with @ravise5

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
rismehta: In that case, we should create a custom serializer and handle this at all places, you can check the way DataRefSerializer is implemented.

bundles/af-core/src/test/resources/form/textinput/exporter-textinput-customized.json
```diff
@@ -4,8 +4,8 @@
   "fieldType": "text-input",
   "name": "abc",
   "visible": false,
-  "description": "dummy",
-  "tooltip": "test-short-description",
+  "description": "&lt;p&gt;dummy&lt;/p&gt;",
```
rismehta, line 7: This would need to be part of crispr spec as wordings and we would have to bump major version of adaptive form property which we emit today, so that existing clients do not break in headless

bundles/af-core/src/test/resources/form/radiobutton/exporter-radiobutton-customized.json
```diff
@@ -5,7 +5,7 @@
   "name": "radiobutton_12345",
   "visible": false,
   "description": "long description",
-  "tooltip": "<p>short description</p>",
+  "tooltip": "&lt;p&gt;short description&lt;/p&gt;",
```
nit23uec, line 8: pls ensure output of model.json API remains unchanged.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
```diff
@@ -227,6 +234,7 @@ public String getHtmlScreenReaderText() {
      */
     @Override
     @Nullable
+    @JsonSerialize(using = EncodeHTMLSerializer.class)
```
nit23uec, line 237: this would affect model.json output also.

TalmizAhmed: ack

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
TalmizAhmed: Implemented using Custom Serializer

bundles/af-core/src/test/resources/form/textinput/exporter-textinput-customized.json
TalmizAhmed: Avoiding this route

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/FormStructureParserImpl.java
```diff
@@ -118,6 +120,7 @@ public String getFormDefinition() {
         FormContainer formContainer = resource.adaptTo(FormContainer.class);
         try {
             ObjectMapper mapper = new ObjectMapper();
+            mapper.registerModule(new SimpleModule().addSerializer(String.class, new AbstractBaseImpl.EncodeHTMLSerializer()));
```
nit23uec, line 123: lets keep the serializer outside AbstractBase.

TalmizAhmed: Done

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
TalmizAhmed: Closing this as this will be unrelated now since the encoding is happening on FormStructParser. Everything else should be working as before without backward incompatibilities.

rismehta (approved):
LGTM

PR #1269, opened 2024-06-11
made changes in getId method where currentPage is null in case of resource adaptation

deepprakash345:
When migrating from eds to master. 

- We moved the getFormDefinition Api from formContainer to FormStructParser. In doing so , we adapted the resource to FormContainer object and then generated the json.
- Before since the context was with request object the currentPage variable was initiated from request obj but now this was null resulting in wrong id getting generated in eds json.

cc:- @nit23uec , @jalagari , @rismehta

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
@@ -216,7 +216,8 @@ public class FormContainerImpl extends AbstractContainerImpl implements FormCont
                 return ComponentUtils.getEncodedPath(getPath());
             }
         } else {
-            return super.getId();
+            String pagePath = getParentPagePath();
+            return (StringUtils.isNotBlank(pagePath))? ComponentUtils.getEncodedPath(pagePath) : super.getId();
         }
     }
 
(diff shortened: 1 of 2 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -216,7 +216,8 @@ public String getId() {
                 return ComponentUtils.getEncodedPath(getPath());
             }
         } else {
-            return super.getId();
+            String pagePath = getParentPagePath();
```
nit23uec, line 219: We need to handle it for forms in sites use case also when curentPage is null.

deepprakash345: Hi @nit23uec , in case of form in sites scenario if currentpage is null, this else block will find the nearest ancestor page which is site page hosting the form and will generate an id on the basis of page.

Am i missing anything.

nit23uec: That is incorrect. Since we want to have the path till form container for this case.

nit23uec: It's better if make id generation here independent of currentPage altogether.

deepprakash345: but @nit23uec , in that case we will end up generating two different ids. In normal aem case, when currentPage is not null, the id generated will be of pagePath. In eds case, when currentPage is null, the id generated will be till guidecontainer path.

rismehta: The id generation logic does the following today,
1. For standalone forms, it is path to the form
2. If forms is used inside SITES, its path to the form container
3. When forms is used inside embed component, the same logic is being improved by @rajatkhurana-adobe 

Can we have unit test cases for all these use-cases in this CL ? ID logic is dependent on the current page also.

> It's better if make id generation here independent of currentPage altogether.

Its not a trivial change to do right now, since there are many dependent modules.

rismehta (changes requested):
check comments

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
nit23uec: > Its not a trivial change to do right now, since there are many dependent modules.
What I'm proposing is: we already have a way to find out the parent page path of a resource. we can use the same and generate id based on whether its a standalone form or form authored in sites. the id would still correspond to page path in case of standalone form and form container i case of form in sites.

deepprakash345: made the changes as suggested in the review

nit23uec (approved):

deepprakash345:
> check comments

HI @rismehta , the if block is getting tested by the getId unit test and the else block is getting tested from cypress test for page v2.

rismehta (approved):

PR #1274, opened 2024-06-18
few properties were not getting updated via rules on form initialization

barshat7:
## Description

Some Properties like 'required' 'description' and 'label' were not getting updated when set via rule 'When Form is Initialized'.
Further, the 'description' was never getting updated as if no description was authored initially then the description div was explicitly ignored in the Sightly. As a fix, included the long description div (which is always hidden by default until question mark icon is clicked), and marked the questionMark div as hidden unless we see that the description is present.

## Types of changes

## Checklist:

### DIFF
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/longDescription.html
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/longDescription.html
@@ -1,5 +1,5 @@
 <template data-sly-template.longDescription="${@ componentId, longDescription, bemBlock}">
-    <div class="${bemBlock}__longdescription" id="${componentId}__longdescription" data-sly-test="${longDescription}" aria-live="polite">
-      ${longDescription @context='html'}
-    </div>
+  <div class="${bemBlock}__longdescription" id="${componentId}__longdescription" style="${wcmmode.edit && longDescription ? '' : 'display:none;'} aria-live="polite">
+  ${longDescription @context='html'}
+  </div>
 </template>
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/longDescription.html
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/longDescription.html
@@ -1,5 +1,5 @@
 <template data-sly-template.longDescription="${@ componentId, longDescription, bemBlock}">
-  <div class="${bemBlock}__longdescription" id="${componentId}__longdescription" style="${wcmmode.edit && longDescription ? '' : 'display:none;'} aria-live="polite">
-  ${longDescription @context='html'}
+  <div class="${bemBlock}__longdescription" id="${componentId}__longdescription" data-sly-test="${longDescription}" aria-live="polite">
+    ${longDescription @context='html'}
   </div>
 </template>
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/questionMark.html
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/questionMark.html
@@ -1,4 +1,4 @@
 <template data-sly-template.questionMark="${@ componentId, longDescription, bemBlock}">
-  <button class="${bemBlock}__questionmark" style="${longDescription ? '' : 'display:none;'} title="Help text">
+  <button class="${bemBlock}__questionmark" data-sly-test="${longDescription}" title="Help text">
   </button>
 </template>
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/questionMark.html
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/questionMark.html
@@ -1,4 +1,4 @@
 <template data-sly-template.questionMark="${@ componentId, longDescription, bemBlock}">
-  <button class="${bemBlock}__questionmark" data-sly-test="${longDescription}" title="Help text">
+  <button class="${bemBlock}__questionmark" style="${longDescription ? '' : 'display:none;'} title="Help text">
   </button>
 </template>
--- a/ui.frontend/src/view/FormFieldBase.js
+++ b/ui.frontend/src/view/FormFieldBase.js
@@ -398,8 +398,8 @@ class FormFieldBase extends FormField {
      * @private
      */
     #showHideLongDescriptionDiv(show) {
-        if (this.description) {
-            this.toggleAttribute(this.description, show, Constants.DATA_ATTRIBUTE_VISIBLE, false);
+        if (this.getDescription()) {
+            this.toggleAttribute(this.getDescription(), show, Constants.DATA_ATTRIBUTE_VISIBLE, false);
         }
     }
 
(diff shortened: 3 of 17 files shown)


ui.tests/test-module/libs/support/index.js
```diff
@@ -111,6 +111,11 @@ Cypress.on('uncaught:exception', (err, runnable) => {
         return false;
     }
 
+    // this error can come if properties are set via rule editor on form initialization (Since field are yet not initialized completely)
+    if (err.message.includes('some error occurred while triggering event on guideBridge')) {
```
rismehta, line 115: The test case should fail, why is this part of whitelist ? Please remove this

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -332,25 +335,29 @@ class FormFieldBase extends FormField {
...
+                panelName,
+                fieldQualifiedName,
+                ...(originalEventPayload?.prevValue !== undefined ? { prevText: originalEventPayload?.prevValue } : {}),
+                ...(originalEventPayload?.currentValue !== undefined ? { newText: originalEventPayload?.currentValue } : {}),
+                ...((typeof originalEventPayload === 'object' && originalEventPayload !== null) ? originalEventPayload : {})
+            };
+            const formContainerPath = this.formContainer.getPath();
+            window.guideBridge.trigger(eventType, eventPayload, formContainerPath);
+        } catch (e) {
+            console.log('some error occurred while triggering event on guideBridge', e);
```
rismehta, line 359: Why did you add a try catch here, ideally the code should break. Can you describe the use-case

rismehta (changes requested):
check comments

ui.tests/test-module/specs/tabsontop/tabsontop.runtime.spec.js
```diff
@@ -263,13 +263,13 @@ describe("setFocus on the first invalid field on submission", () => {
     });
 
     it("check if invalid field gets focus on submit button click", () => {
-           cy.get('button').should('be.visible').click().then(() => {
+           cy.get('button').last().should('be.visible').click().then(() => {
```
rismehta, line 266: Please don't change anything in an existing test case, unless there is backward incompatible change

barshat7: Previously we did not include the questionMark div if long description was not set, but since the setProperty can be called at runtime and even if long description was not set we need to set it in runtime, then we need to have this div. As per this change https://github.com/adobe/aem-core-forms-components/pull/1274/files#:~:text=%3Cbutton%20class%3D%22%24%7BbemBlock%7D__questionmark%22%20style%3D%22%24%7BlongDescription%20%3F%20%27%27%20%3A%20%27display%3Anone%3B%27%7D%20title%3D%22Help%20text%22%3E

Now in the test cases the selectors were wrong as it was assuming that there would only be one button in the form, so I had to update the selectors. 
Not sure if this would be backward incompatible though

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -602,7 +608,17 @@ class FormFieldBase extends FormField {
...
-            this.description.querySelector("p").innerHTML = description;
+            if (description) {
+                if (!this.description.querySelector("p")) {
+                    // If the description is updated via rule then it might not have <p> tags
+                    this.description.appendChild(document.createElement('p'));
+                }
+                this.description.querySelector("p").innerHTML = description;
+                const questionMarkDiv = this.getQuestionMarkDiv();
+                 if (questionMarkDiv) {
+                     questionMarkDiv.style.display = "block";
```
rismehta, line 619: We should not be updating styles via code, this might override the theme, and might conflict. What is the use-case to override style

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/questionMark.html
```diff
@@ -1,4 +1,4 @@
 <template data-sly-template.questionMark="${@ componentId, longDescription, bemBlock}">
-  <button class="${bemBlock}__questionmark" data-sly-test="${longDescription}" title="Help text">
+  <button class="${bemBlock}__questionmark" style="${longDescription ? '' : 'display:none;'} title="Help text">
```
rismehta, line 2: styles need to come via theme, code should not add inline style, it would override other styles provided by customer. This looks incorrect to me

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/fieldTemplates/longDescription.html
```diff
@@ -1,5 +1,5 @@
 <template data-sly-template.longDescription="${@ componentId, longDescription, bemBlock}">
-    <div class="${bemBlock}__longdescription" id="${componentId}__longdescription" data-sly-test="${longDescription}" aria-live="polite">
-      ${longDescription @context='html'}
-    </div>
-</template>
\ No newline at end of file
+  <div class="${bemBlock}__longdescription" id="${componentId}__longdescription" style="${wcmmode.edit && longDescription ? '' : 'display:none;'} aria-live="polite">
```
rismehta, line 2: Why do we need inline style

ui.frontend/src/view/FormFieldBase.js
barshat7: The requirement is that if long description is not set during authoring then the questionMark div and description div should be present in DOM but hidden. And if by any rule the description is set in runtime, then they should be visible. I guess we could do that by adding 'data-cmp-visible' attribute instead

ui.tests/test-module/specs/tabsontop/tabsontop.runtime.spec.js
rismehta: But this would pollute the HTML with unncessary markup for each field, thus can impact HTML parsing even for use-cases where description is not set at runtime, which is incorrect.

Ideally, at runtime, we should generate the markup if description is set at runtime.

rismehta (changes requested):
check comments

ui.frontend/src/view/FormFieldBase.js
rismehta: yes, please check this comment on the overall design, https://github.com/adobe/aem-core-forms-components/pull/1274#discussion_r1654523515

rismehta:
Also, do we have a customer who has asked for this change ? Can you link the JIRA in the PR

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -598,16 +604,45 @@ class FormFieldBase extends FormField {
...
+        if (descriptionText !== undefined) {
+            if (this.getDescription()) {
+                if (!this.getDescription().querySelector("p")) {
+                    // If the description is updated via rule then it might not have <p> tags
+                    this.getDescription().appendChild(document.createElement('p'));
+                }
+            } else {
+                this.#addDescriptionInRuntime();
+            }
+            this.getDescription().querySelector("p").innerHTML = descriptionText;
```
rismehta, line 619: You would need to sanitize this HTML, we already do this for enumNames,

```
window.DOMPurify ?  window.DOMPurify.sanitize(descriptionText) : descriptionText
```

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -598,16 +604,45 @@ class FormFieldBase extends FormField {
...
-     */
-    updateDescription(description) {
-        if (this.description) {
-            this.description.querySelector("p").innerHTML = description;
-        } else {
-            //TODO: handle the case when description is not present initially.
+     * @param {string} descriptionText - The description.
+     */
+    updateDescription(descriptionText) {
+        if (descriptionText !== undefined) {
```
rismehta, line 610:
```  
  if (typeof descriptionText !== 'undefined') {
        const descriptionElement = this.getDescription();
        if (descriptionElement) {
            let pElement = descriptionElement.querySelector("p");
            if (!pElement) {
                // If the description is updated via rule then it might not have <p> tags
                pElement = document.createElement('p');
                descriptionElement.appendChild(pElement);
            }
            pElement.textContent = descriptionText;
        } else {
            this.#addDescriptionInRuntime(descriptionText);
        }
    }
    ```

File: ui.frontend/src/view/FormFieldBase.js [hunk 9]
```diff
@@ -598,16 +604,45 @@ class FormFieldBase extends FormField {
...
+                }
+            } else {
+                this.#addDescriptionInRuntime();
+            }
+            this.getDescription().querySelector("p").innerHTML = descriptionText;
         }
     }
 
+    #addDescriptionInRuntime() {
+        console.log('adding description in runtime')
```
rismehta, line 624: Remove the console log

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -598,16 +604,45 @@ class FormFieldBase extends FormField {
...
+        // add question mark icon
+        const bemClass = Array.from(this.element.classList).filter(bemClass => !bemClass.includes('--'))[0];
```
rismehta, line 626: lets make the code NPE safe

```
 // add question mark icon
    const bemClass = Array.from(this.element.classList).find(bemClass => !bemClass.includes('--'));
    const labelContainer = this.element.querySelector(`.${bemClass}__label-container`);
    if (labelContainer) {
        const qmButton = document.createElement('button');
        qmButton.className = `${bemClass}__questionmark`;
        qmButton.title = 'Help text';
        labelContainer.appendChild(qmButton);
    } else {
        console.error('Label container not found');
        return;
    }

    // add description div
    const descriptionDiv = document.createElement('div');
    descriptionDiv.className = `${bemClass}__longdescription`;
    descriptionDiv.id = `${this.getId()}__longdescription`;
    descriptionDiv.setAttribute('aria-live', 'polite');
    descriptionDiv.setAttribute('data-cmp-visible', 'false');
    
    const pElement = document.createElement('p');
    if (typeof descriptionText !== 'undefined') {
        pElement.textContent = descriptionText;
    }
    descriptionDiv.appendChild(pElement);

    const errorDiv = this.getErrorDiv();
    if (errorDiv) {
        this.element.insertBefore(descriptionDiv, errorDiv);
    } else {
        console.error('Error div not found');
        return;
    }
```

rismehta (changes requested):
check comments

rismehta (approved):

PR #1293, opened 2024-06-27
FORMS-15150 Runtime client libraries are loaded in authoring for tabs…

dmaurya929:
… and accordion

## Types of changes

## Checklist:


ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/accordion/v1/accordion/clientlibs/commons/js/common.js
```diff
@@ -0,0 +1,261 @@
...
+                    panel.classList.add(this.constructor.cssClasses.panel.hidden);
+                    panel.classList.remove(this.constructor.cssClasses.panel.expanded);
+                    panel.setAttribute("aria-hidden", true);
+                }
+            }
+        }
+    }
+
+    window.CQ = window.CQ || {};
+    window.CQ.FormsCoreComponents = window.CQ.FormsCoreComponents || {};
```
rismehta, line 257: `window.CQ.FormsCoreComponents` we use this namespace only for authoring. We have not had use-cases of commons yet, so let's namespace this separately mentioning the commons.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/accordion/v1/accordion/clientlibs/commons/js/common.js
```diff
@@ -0,0 +1,261 @@
...
+                item: {
+                    expanded: "data-cmp-expanded"
+                }
+            };
+
+            constructor(params) {
+                super(params);
+            }
+
+            _getCachedItems() {
```
rismehta, line 65: Please use the new way to mention private methods using #

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/accordion/v1/accordion/clientlibs/editorframe/.content.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
+    jcr:primaryType="cq:ClientLibraryFolder"
+    allowProxy="{Boolean}true"
+    categories="[core.forms.components.accordion.v1.editorframe]"
```
rismehta, line 5: We should add this as part of editor hook. This way we don't need to include it in the component. Client libraries should always be part of header/footer for better performance

rismehta (changes requested):
please add test cases for this change

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/accordion/v1/accordion/clientlibs/commons/js/common.js
dmaurya929: This is base class for `Accordion`, if using `#` then child class would not be able to access. So changed to `_`  just to represent it as private.

rismehta (changes requested):
I had done some changes in tabs, can you incorporate that in this PR, else that use-case would break. Also, could you fix the wizard / accordion issue too.

dmaurya929:
> I had done some changes in tabs, can you incorporate that in this PR, else that use-case would break. Also, could you fix the wizard / accordion issue too.

Updated the PR, also issue is not reproducible in wizard / accordion so doesn't require any changes there.
cc: @rismehta

rismehta:
Please re-base the PR by resolving the conflicts

PR #1323, opened 2024-07-18
image.1.json is returning model.json, as a fix revert PR-243 and adde…

barshat7:
…d extensions

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/StaticImageGETServlet.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/StaticImageGETServlet.java
@@ -21,7 +21,6 @@ import java.io.InputStream;
 import javax.jcr.Property;
 import javax.jcr.RepositoryException;
 import javax.servlet.Servlet;
-import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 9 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/StaticImageGETServlet.java
```diff
@@ -49,27 +46,13 @@
     service = { Servlet.class },
     property = {
         "sling.servlet.resourceTypes=" + FormConstants.RT_FD_FORM_IMAGE_V1,
+        "sling.servlet.extensions=jpg",
+        "sling.servlet.extensions=png",
+        "sling.servlet.extensions=gif",
         "sling.servlet.methods=GET"
     })
 public class StaticImageGETServlet extends AbstractImageServlet {
```
rismehta, line 54: Why do we even need this servlet ? We changed the header component to use the image from WCM components. ideally, this image should also extend WCM component and we should just be adding fieldType like title or plain-text.

rismehta (changes requested):
check comments

rismehta (changes requested):
check comments

rismehta (approved):
Approving, since we need this for feature parity and for backward compatibility.

PR #1348, opened 2024-08-08
Fixing Min/Max dialog fields for number input, when the type is decimal

vdua:
- randomly chosen step as 0.01 for decimal, but it can be author defined

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/package-info.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/package-info.java
@@ -35,7 +35,7 @@
  * </p>
  */
 
-@Version("5.5.4") // aligning this with release/650 since af2-rest-api is compiled with 5.2.0 in release/650
+@Version("5.6.4") // aligning this with release/650 since af2-rest-api is compiled with 5.2.0 in release/650
 package com.adobe.cq.forms.core.components.models.form;
 
 import org.osgi.annotation.versioning.Version;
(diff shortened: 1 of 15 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/package-info.java
```diff
@@ -35,7 +35,7 @@
  * </p>
  */
 
-@Version("5.5.4") // aligning this with release/650 since af2-rest-api is compiled with 5.2.0 in release/650
+@Version("5.6.4") // aligning this with release/650 since af2-rest-api is compiled with 5.2.0 in release/650
```
vdua, line 38: the comment doesn't make sense here.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/numberinput/v1/numberinput/clientlibs/editor/js/editDialog.js
```diff
@@ -52,6 +52,37 @@
...
+                maxField.step = "1";
+                // Reset value if it is a decimal
+                if (minInput.value.includes('.')) {
+                    minInput.value = Math.floor(minInput.value);
+                    minInput.setAttribute('value', minInput.value);
+                }
+                if (maxInput.value.includes('.')) {
+                    maxInput.value = Math.floor(maxInput.value);
+                    maxInput.setAttribute('value', maxInput.value);
+                }
```
vdua, line 80: can we add a test case for this use case where changing from number to integer, changes the min/max constraint as well

PR #1365, opened 2024-08-29
Screen Reader Text Support in core components

pavi41:
## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
```diff
@@ -182,12 +182,12 @@ public String getScreenReaderText() {
         if (AssistPriority.LABEL.equals(assistPriority)) {
             Label label = getLabel();
             if (label != null) {
-                screenReaderText = "$label.$value";
+                screenReaderText = label.getValue();
```
rismehta, line 185: Since label can be dynamic, thats why this was added as an expression. Ideally, this needs to be evaluated by af-core dynamically at runtime.

rismehta (changes requested):
check comment. we need to fix this in af-core

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -222,10 +223,27 @@ class FormFieldBase extends FormField {
...
     }
 
+    #syncAriaLabel() {
+        let widgetElement = typeof this.getWidget === 'function' ? this.getWidget() : null;
+        let widgetElements = typeof this.getWidgets === 'function' ? this.getWidgets() : null;
+        widgetElement = widgetElements || widgetElement;
+
+        if (widgetElement && this.getModel() && this.getModel().screenReaderText) {
+            // Create a temporary DOM element to strip HTML tags
+            const tempElement = document.createElement('div');
```
rismehta, line 238: Let;s not use DOM API's for this, we already have DOMPurify for this

```
#syncAriaLabel() {
    const widgetElement = this.getWidget?.() || this.getWidgets?.();
    const model = this.getModel?.();

    if (widgetElement && model?.screenReaderText) {
        // Use DOMPurify to sanitize and strip HTML tags
        const screenReaderText = DOMPurify.sanitize(model.screenReaderText, { RETURN_TRUSTED_TYPE: true });
        widgetElement.setAttribute('aria-label', screenReaderText);
    }
}
```

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractBaseImpl.java
rismehta: Can you open a backlog ticket to fix this ? For now, let's keep the value as static

ui.frontend/src/view/FormFieldBase.js
pavi41: Implemented!

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -222,10 +223,27 @@ class FormFieldBase extends FormField {
...
+            tempElement.innerHTML = this.getModel().screenReaderText;
+            // Get the text content without HTML tags
+            const screenReaderText = tempElement.textContent || tempElement.innerText || '';
+            widgetElement.setAttribute('aria-label', screenReaderText);
```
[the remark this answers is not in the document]
pavi41: Discussed with Devendra!

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -222,10 +223,27 @@ class FormFieldBase extends FormField {
...
 
+    #syncAriaLabel() {
+        let widgetElement = typeof this.getWidget === 'function' ? this.getWidget() : null;
+        let widgetElements = typeof this.getWidgets === 'function' ? this.getWidgets() : null;
+        widgetElement = widgetElements || widgetElement;
+        const model = this.getModel?.();    
+    
+        if (widgetElement && model?.screenReaderText) {
+            // Use DOMPurify to sanitize and strip HTML tags
+            const screenReaderText = window.DOMPurify.sanitize(model.screenReaderText, { ALLOWED_TAGS: [] });
```
rismehta, line 239: Can you also add a NPE check on window.DOMPurity

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -608,16 +627,10 @@ class FormFieldBase extends FormField {
...
             let descriptionElement = this.getDescription();
             if (descriptionElement) {
-                let pElement = descriptionElement.querySelector("p");
-                if (!pElement) {
-                    // If the description is updated via rule then it might not have <p> tags
-                    pElement = document.createElement('p');
-                    descriptionElement.appendChild(pElement);
-                }
-                pElement.textContent = sanitizedDescriptionText;
+                descriptionElement.innerHTML = sanitizedDescriptionText;
```
rismehta, line 633: Can we optimize this ? if the content is already the same, let's not update the HTML. This would be a NO-OP operation in few cases which needs to be handled

rismehta (changes requested):
check comments

ui.frontend/src/view/FormFieldBase.js
```diff
@@ -617,7 +637,7 @@ class FormFieldBase extends FormField {
                     pElement = document.createElement('p');
                     descriptionElement.appendChild(pElement);
                 }
-                pElement.textContent = sanitizedDescriptionText;
+                pElement.innerHTML = sanitizedDescriptionText;
```
rismehta, line 640: I think we need to improve this logic as follows, could you validate if this would work for all use-cases, including the migration bug which is discussed over slack


```
const sanitizedDescriptionText = window.DOMPurify ? window.DOMPurify.sanitize(descriptionText) : descriptionText;
let descriptionElement = this.getDescription();

if (descriptionElement) {
    // Check if the content inside the descriptionElement needs updating
    let currentTextContent = descriptionElement.innerText.trim();

    if (currentTextContent === sanitizedDescriptionText) {
        // No update needed if the text content already matches
        return;
    }

    // Find the existing <p> element
    let pElement = descriptionElement.querySelector('p');

    if (!pElement) {
        // If no <p> tag exists, create one and set it as the content
        pElement = document.createElement('p');
        descriptionElement.innerHTML = ''; // Clear existing content
        descriptionElement.appendChild(pElement);
    }

    // Update the <p> element's content with sanitized content
    pElement.innerHTML = sanitizedDescriptionText;
}


```

rismehta (changes requested):
check comments

rismehta (approved):

PR #1368, opened 2024-08-30
fix: adding exception for accessibility failing in the pipeline

prateek1497:
## Types of changes

## Checklist:


it/content/src/main/content/jcr_root/content/forms/af/core-components-it/samples/accessibility/.content.xml
```diff
@@ -60,7 +60,7 @@
                     sling:resourceType="forms-components-examples/components/form/panelcontainer"
                     enabled="{Boolean}true"
                     fieldType="panel"
-                    hideTitle="false"
+                    hideTitle="true"
```
prateek1497, line 63: added this because: because the panel title was coming before: h1 tag

https://dequeuniversity.com/rules/axe/4.1/page-has-heading-one

Generally, it is a best practice to ensure that the beginning of a page's main content starts with a h1 element, and also to ensure that the page contains only one h1 element.

PR #1409, opened 2024-09-17
UE: Binding each core component's fieldType to its respective sling model

TalmizAhmed:
## Description
Each core component no longer has to mandatorily provide a fieldType in JCR, it will be inferred in the sling model and provided in the json.

## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/ButtonImpl.java
```diff
@@ -87,6 +88,14 @@ public String getButtonType() {
         return buttonType;
     }
 
+    @Override
+    public String getFieldType() {
```
rismehta, line 92: We could move the logic to compute field type in base class as a function, and re-use it everywhere else. Also, we can use the StringUtils.isBlank check for fieldTypeJcr to check empty strings as well

rismehta (changes requested):
check comments

rismehta (changes requested):
Can you move the spec to 0.14.1 version and also update the new fields types here as per 0.14.1

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/ButtonImpl.java
TalmizAhmed: Used StringUtils.IsNotEmpty to check for empty and null

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -216,6 +218,13 @@ public String getFieldType() {
         return fieldType.getValue();
     }
 
+    public String getFieldType(@Nonnull FieldType defaultFieldType) {
```
rismehta, line 221: This should be protected

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -78,9 +78,9 @@ public class AbstractFormComponentImpl extends AbstractComponentImpl implements
     protected String dataRef;
 
     // mandatory property else adapt should fail for adaptive form components
-    @ValueMapValue(name = ReservedProperties.PN_FIELDTYPE)
+    @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL, name = ReservedProperties.PN_FIELDTYPE)
     protected String fieldTypeJcr;
-    private FieldType fieldType;
+    protected FieldType fieldType;
```
nit23uec, line 83: This can be kept private, I believe?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FieldType.java
```diff
@@ -35,7 +35,10 @@ public enum FieldType {
     PANEL("panel"),
     FORM("form"),
     CHECKBOX_GROUP("checkbox-group"),
-    IMAGE("image");
+    IMAGE("image"),
+    TELEPHONE("tel"),
```
nit23uec, line 39: We would want to support password and range also in eds, but may be not in cc for the starting. Either we should the sling model or give a static mapping for such components.

rismehta: password is already being worked upon as part of this PR, https://github.com/adobe/aem-core-forms-components/pull/1316

bundles/af-core/src/test/resources/schema/0.14.0/adaptive-form.schema.json
```diff
@@ -811,7 +811,9 @@
             "rules",
             "screenReaderText",
             "tooltip",
-            "visible"
+            "visible",
+            "type",
+            "buttonType"
```
rismehta, line 816: Why have we added buttonType here ? When was this added as part of spec ?

rismehta (changes requested):
check comments

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FieldType.java
TalmizAhmed: Without having it defined now would mean that it would default to text-input. Which is compatible for now until the password core component is done.

bundles/af-core/src/test/resources/schema/0.14.0/adaptive-form.schema.json
TalmizAhmed: Ideally shouldn't buttonType be a part of the spec?
Did this for the test case for JSON export, changing this for a simpler check now.

rismehta (approved):

PR #1415, opened 2024-09-19
[Draft] FORMS-16201 Register auto save in publish(disabled) mode only

girotraapankaj:
## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -379,4 +379,10 @@ public String getName() {
         return FormContainer.super.getName();
     }
 
+    @JsonIgnore
+    @Override
+    public AutoSaveConfiguration getAutoSaveConfig() {
+        return autoSaveConfig;
+    }
```
devgurjar, line 386: Do we need to document it for headless as well?

girotraapankaj: yes, for headless this will be part of properties json. Right now, we need to use this in sightly so added getter but ignored from json

PR #1454, opened 2024-10-21
UE: Supporting a new CRX property for multi select dropdown's default values

TalmizAhmed:
## Description
https://jira.corp.adobe.com/browse/FORMS-16246

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
@@ -74,6 +74,7 @@ public final class ReservedProperties {
     public static final String PN_PLACEHOLDER = "placeholder";
     public static final String PN_READ_ONLY = "readOnly";
     public static final String PN_DEFAULT_VALUE = "default";
+    public static final String PN_MULTI_DEFAULT_VALUES = "multiDefaultValues";
     public static final String PN_FORMAT = "format";
     public static final String PN_DISPLAY_FORMAT = "displayFormat";
     public static final String PN_EDIT_FORMAT = "editFormat";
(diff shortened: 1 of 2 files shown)


rismehta (changes requested):
Please add test cases. Shouldn't the new property be prefixed with fd: ?

TalmizAhmed:
> Please add test cases. Shouldn't the new property be prefixed with fd: ?

@rismehta I don't think so, Since we do not have special logic for its handling, It acts like a new OOTB property that is a part of the reserved properties list.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
```diff
@@ -74,6 +74,7 @@ private ReservedProperties() {
     public static final String PN_PLACEHOLDER = "placeholder";
     public static final String PN_READ_ONLY = "readOnly";
     public static final String PN_DEFAULT_VALUE = "default";
+    public static final String PN_MULTI_DEFAULT_VALUES = "multiDefaultValues";
```
nit23uec, line 77: lets capture this change in our migration wiki also.

nit23uec (approved):

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
rismehta, line 77: This property should be prefixed with fd: right ?

TalmizAhmed: @rismehta Replied to this earlier: https://github.com/adobe/aem-core-forms-components/pull/1454#issuecomment-2426015289

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractOptionsFieldImpl.java
```diff
@@ -50,6 +50,10 @@ public abstract class AbstractOptionsFieldImpl extends AbstractFieldImpl impleme
     @Nullable
     protected String[] enumNames;
 
+    @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL, name = ReservedProperties.PN_MULTI_DEFAULT_VALUES)
+    @Nullable
+    protected Object[] multiDefaultValues;
```
rismehta, line 55: Can we add a code comment here, that this property would only exist for forms created in UE, also add reasoning on why this is done

rismehta (changes requested):
check comments

PR #1455, opened 2024-10-21
FORMS-16943: added dateToDaysSinceEpoch method

arunattri26:
## Types of changes

## Checklist:


ui.frontend/package-lock.json
```diff
@@ -70,9 +70,9 @@
       }
     },
     "node_modules/@aemforms/af-custom-functions": {
-      "version": "1.0.10",
-      "resolved": "https://registry.npmjs.org/@aemforms/af-custom-functions/-/af-custom-functions-1.0.10.tgz",
-      "integrity": "sha512-n3w9tHkJOI5ISVYAK2cCi5k/oTu3rGgByDmMIgOH1+Ry4mL9nM3cxBTKEkPF8Y8JiKF1aUHIKM+MeP6u5PiiUA=="
+      "version": "1.0.12",
+      "resolved": "https://artifactory.corp.adobe.com/artifactory/api/npm/npm-aem-release/@aemforms/af-custom-functions/-/af-custom-functions-1.0.12.tgz",
```
rismehta, line 74: This URL should points to npmjs

arunattri26: Done.

rismehta (changes requested):
Can you also add a test case for this use-case in this PR via visual rule editor ?

arunattri26:
> Can you also add a test case for this use-case in this PR via visual rule editor ?

I will add it after it is merged. Since rule-editor would expect it to be present else the rule will fail. So we can't release rule-editor without this function being present in core-components.

rismehta:
> > Can you also add a test case for this use-case in this PR via visual rule editor ?
> 
> I will add it after it is merged. Since rule-editor would expect it to be present else the rule will fail. So we can't release rule-editor without this function being present in core-components.

The rule editor shouldn't rely on the core component version, as it may also fail with older versions. Please include test cases in this PR and allow them to fail until the new version of the rule editor SPA is released.

arunattri26:
> > > Can you also add a test case for this use-case in this PR via visual rule editor ?
> > 
> > 
> > I will add it after it is merged. Since rule-editor would expect it to be present else the rule will fail. So we can't release rule-editor without this function being present in core-components.
> 
> The rule editor shouldn't rely on the core component version, as it may also fail with older versions. Please include test cases in this PR and allow them to fail until the new version of the rule editor SPA is released.

Done. I have added the tests for latest-addon and only my authoring test is failing.

rismehta:
@arunattri26 One of the test case is failing, please check and fix it

arunattri26:
> @arunattri26 One of the test case is failing, please check and fix it

@rismehta , I checked it is failing due to old quickstart we are using in tests which does not have the latest exp-editor. After updating the QS with latest devrelease/sites-rotary branch it will get resolved.

PR #1461, opened 2024-10-23
update url to register custom function via guidebridge config

barshat7:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/HTTPAPILayer.js
+++ b/ui.frontend/src/HTTPAPILayer.js
@@ -153,7 +153,7 @@ class HTTPAPILayer {
         if (customFunctionConfig && customFunctionConfig.length === 1) {
                 const getCustomFunctionUrl = customFunctionConfig[0].fn;
                 if (typeof getCustomFunctionUrl === 'function') {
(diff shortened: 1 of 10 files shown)


ui.frontend/src/HTTPAPILayer.js
```diff
@@ -148,7 +148,18 @@ class HTTPAPILayer {
      * @returns {Promise<Object>} - A Promise that resolves to the custom function configuration.
      */
     static async getCustomFunctionConfig(formId) {
-        return await this.getJson(Constants.API_PATH_PREFIX + "/customfunctions/" + formId);
+        let customFunctionsUrl;
```
rismehta, line 151: This appears quite cryptic and hard to read.

A better approach would be to add this as a custom property on the server side, similar to how we handle action and data URLs. Customers could then modify the server-side Sling model as they do for other URLs.

However, this wouldn’t address the use case where we access AEM through hardcoded URLs on the client side, such as the toggles.json call. We need to consider how to incorporate prefixes for those hardcoded URLs.

Other URLs that come from the server or HTML should work fine with the Sling rewriter approach.

rismehta (changes requested):
check comments

PR #1469, opened 2024-10-28
custom function registration using action url from backend

barshat7:
custom function registration using action url from backend

## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/FormContainer.java
```diff
@@ -373,4 +373,26 @@ default void visit(Consumer<ComponentExporter> callback) throws Exception {}
...
     }
+
+    /**
+     * Returns the url from where the custom functions should be registered
+     * 
+     * @return custom function registration url
+     * @since com.adobe.cq.forms.core.components.models.form 5.7.6
+     */
+    @JsonIgnore
+    default String getCustomFunctionAction() {
```
rismehta, line 384: Please refer to the implementations of getAction and getDataUrl; there should be no property for this in JCR. For getPrefillAction, let's reuse the existing data URL implementation and rename it to getCustomFunctionURL.

ui.frontend/src/utils.js
```diff
@@ -309,14 +327,15 @@ class Utils {
             } else {
                 const _formJson = await HTTPAPILayer.getFormDefinition(_path, _pageLang);
                 console.debug("fetched model json", _formJson);
-                await this.registerCustomFunctions(_formJson.id);
```
rismehta, line 312: We could improve the existing function implementation to incorporate the following logic,
Check if fd:customFunctionsUrl exists, if yes, use it, else fallback to the existing logic

let's not create a new method

rismehta (changes requested):
check comments

ui.frontend/src/utils.js
barshat7: The intention to create a new function was not to change the signature of the existing one. I see the existing function as exposed and documented.

rismehta: @barshat7 You were right, this function is a public API, https://opensource.adobe.com/aem-core-forms-components/module-FormView-HTTPAPILayer.html

Can we deprecate this API and create a new API with proper JS doc annotation in each API

rismehta (changes requested):
check comments

rismehta (changes requested):
check comments

ui.frontend/src/utils.js
```diff
@@ -286,6 +287,30 @@ class Utils {
         }
     }
 
+    /**
+     * Registers custom functions from clientlibs.
+     * @param {object} formJson - The form ID.
```
rismehta, line 292: Please update js doc since these are auto published.

PR #1489, opened 2024-12-02
Add stepped class in wizard/tabs

pavi41:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/view/FormTabs.js
+++ b/ui.frontend/src/view/FormTabs.js
@@ -286,7 +286,27 @@ class FormTabs extends FormPanel {
      */
     navigate(tabId) {
         this.#_active = tabId;
+        this.addSteppedClass(tabId);
         this.#refreshActive();
+    }
+
+
+    addSteppedClass(tabId) {
+        var tabs = this.#getCachedTabs();
+        var tabpanels= this.#getCachedTabPanels();
+        const activeTabId = this.getActiveTabId(tabs); 
+        const activeTabElement = this.#getTabNavElementById(activeTabId);
+        if (activeTabElement.classList.contains(this.#_selectors.active.tab)) {
+            activeTabElement.classList.add(this.#_selectors.stepped.tab);
+
+            const correspondingPanel = Array.from(tabpanels).find(panel =>
+                panel.getAttribute("aria-labelledby") === activeTabElement.id
+            );
+    
+            if (correspondingPanel) {
+                correspondingPanel.classList.add(this.#_selectors.stepped.tabpanel);
+            }
+        }
     }
 
     /**
@@ -384,6 +404,14 @@ class FormTabs extends FormPanel {
             this.#cacheElements(this._elements.self);
             var repeatedTabPanel = this.#getTabPanelElementById(childView.id + this.#tabPanelIdSuffix);
             repeatedTabPanel.setAttribute("aria-labelledby", childView.id + this.#tabIdSuffix);
+            const steppedTabClass = Array.from(navigationTabToBeRepeated.classList).find(cls => cls.includes('--stepped'));
+            if (steppedTabClass) {
+                navigationTabToBeRepeated.classList.remove(steppedTabClass);
+            }
+            const steppedTabpanelClass = Array.from(repeatedTabPanel.classList).find(cls => cls.includes('--stepped'));
+            if (steppedTabpanelClass) {
+                repeatedTabPanel.classList.remove(steppedTabpanelClass);
+            }
             this.#bindEventsToTab(navigationTabToBeRepeated.id);
             this.#refreshActive();
             if (childView.getInstanceManager().getModel().minOccur != undefined && childView.getInstanceManager().children.length > childView.getInstanceManager().getModel().minOccur) {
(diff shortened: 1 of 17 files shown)


ui.frontend/src/view/FormTabs.js
```diff
@@ -116,6 +116,11 @@ class FormTabs extends FormPanel {
                         tabs[i].setAttribute(Constants.TABINDEX, "0");
                         tabs[i].setAttribute(Constants.ARIA_CURRENT, "true");
                     } else {
+                         // If the panel/tab was previously active, mark it as stepped
+                         if (tabpanels[i].classList.contains(this.#_selectors.active.tabpanel)) {
```
rismehta, line 120: Can you validate the following use-cases,
1. If tabs are dynamically added/removed (via repeatable settings), this indexing might break. Please add test cases for this use-case

for all layout types

rismehta (changes requested):
check comment

ui.frontend/src/view/FormTabs.js
sakshi-arora1: since it is being directly added to HTML, and is not something which is repopulated, so this might work.
Whenever we introduce lazy loading, at that time this will break.
Nevertheless, we should have a repeatable test cases.

ui.tests/test-module/specs/wizard/wizard.runtime.repeatability.cy.js
```diff
@@ -232,7 +269,7 @@ describe('visibility of navigation buttons', function () {
         // check when new instance of repeatable panel is added
         // and that new instance gets active & prev and next nav buttons are visible
         cy.get(`#${textInputId}`).find(driverTextInput).focus().clear().type('a').blur().then(() => {
-            getWizardPanelAtIndex(2).should('have.class', wizardPanelActive);
+            getWizardPanelAtIndex(11).should('have.class', wizardPanelActive);
```
rismehta, line 272: When there is no change in collateral, why is this test case changed ?

ui.tests/test-module/specs/wizard/wizard.runtime.repeatability.cy.js
```diff
@@ -151,6 +151,43 @@ describe("Form with Wizard Container", () => {
             getWizardPanels().should('have.length', 6);
         })
     });
+
+    it("After clicking on every tab, if tab-1 is repeated, the repeated instance should not have stepped class", () => {
+        getTabs().should('have.length', 4);
+        getWizardPanels().should('have.length', 4);
```
rismehta, line 157: I don't see code to add/remove instances in between the panel, which was asked as per last review comments

rismehta (changes requested):
check comments

ui.tests/test-module/specs/wizard/wizard.runtime.repeatability.cy.js
```diff
@@ -151,6 +151,43 @@ describe("Form with Wizard Container", () => {
...
+        cy.get(".cmp-adaptiveform-wizard__nav--next").click({force: true});
+        getTabAtIndex(3).should('have.class', 'cmp-adaptiveform-wizard__tab--active');
+        getWizardPanelAtIndex(3).should('have.class', 'cmp-adaptiveform-wizard__wizardpanel--active');
+        getTabAtIndex(2).should('have.class', 'cmp-adaptiveform-wizard__tab--stepped');
+        getWizardPanelAtIndex(2).should('have.class', 'cmp-adaptiveform-wizard__wizardpanel--stepped');
+        getTabAtIndex(1).should('have.class', 'cmp-adaptiveform-wizard__tab--stepped');
+        getWizardPanelAtIndex(1).should('have.class', 'cmp-adaptiveform-wizard__wizardpanel--stepped');
+        getTabAtIndex(0).should('have.class', 'cmp-adaptiveform-wizard__tab--stepped');
+        getWizardPanelAtIndex(0).should('have.class', 'cmp-adaptiveform-wizard__wizardpanel--stepped');
+        cy.get("button").contains("+R1").click().then(() => {
```
pavi41, line 181: @rismehta I have added the instance, you can check it here!

ui.tests/test-module/specs/wizard/wizard.runtime.repeatability.cy.js
pavi41: It was failing on my local, because the panel at index 11 was active, and it was checking at index 2.

rismehta: The test case was working earlier, ideally if an old test case is changed, it is most like due to collateral else this could be a regression. please investigate

ui.tests/test-module/specs/wizard/wizard.runtime.repeatability.cy.js
pavi41: I have added the instances in tabs/wizards, you can check it in the test case as well.

ui.tests/test-module/specs/wizard/wizard.runtime.repeatability.cy.js
pavi41: Done, reverted to old scenario!

PR #1562, opened 2025-03-21
FORMS-17107: Client-side custom function parsing changes

arunattri26:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/utils.js
+++ b/ui.frontend/src/utils.js
@@ -343,6 +343,7 @@ class Utils {
         }
     }
 
+    
     /**
      * For backward compatibility with older data formats of prefill services like FDM.
      * @param {object} prefillJson - The prefill JSON object.
--- a/ui.frontend/src/utils.js
+++ b/ui.frontend/src/utils.js
@@ -301,14 +301,19 @@ class Utils {
             funcConfig = await HTTPAPILayer.getCustomFunctionConfig(formJson.id);
         }
         console.debug("Fetched custom functions: " + JSON.stringify(funcConfig));
-        if (funcConfig && funcConfig.customFunction) {
-            const funcObj = funcConfig.customFunction.reduce((accumulator, func) => {
-                if (window[func.id]) {
+        
+        if (funcConfig && funcConfig.clientSideParsingEnabled) {
+            this.registerCustomFunctionsV3(formJson);
+        } else {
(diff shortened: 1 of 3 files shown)


rismehta (changes requested):
Can we add test case for this.

ui.frontend/src/utils.js
```diff
@@ -385,6 +390,78 @@ class Utils {
         }
     }
 
+    /**
+     * Registers custom functions from form json.
+     * @param {object} formJson - The Sling model exporter representation of the form
+     */
+    static async registerCustomFunctionsV3(formJson) {
```
rismehta, line 397: Can we create a separate utility file specifically for the rule editor, parallel to utils.js? Ideally, this code isn't something customers need to understand, so having a dedicated utility file for the rule editor would help clarify which module owns it.

vdua: Better naming than V3 and also proper documentation on what is the input/output

ui.frontend/src/utils.js
```diff
@@ -385,6 +390,78 @@ class Utils {
...
+        // Extract custom function names from form events
+        const extractFunctionNames = (formJson) => {
+            const functionNames = new Set();
+            
+            // Helper function to extract function names from event strings
+            const extractFromEventString = (eventString) => {
+                // Match patterns like functionName() with optional parameters
+                const functionPattern = /(\w+)\s*\(/g;
```
vdua, line 405: This is problematic and there are edge cases where the field name contain `(`. Not possible through UI, but with multiple ways of creating a form, it may be possible.

I would parse the event string using JSON Formula and then extract out the functions from AST.

arunattri26: As discussed, we are also checking for function definition to be present in the window object before registering a custom functions which should take care of edge case that some form field might have name like abc() and it should not be registered as custom function in this case.
Also, using JSON formula for parsing and extracting function name might have performance impact on the form rendering as well. So for now we are good with this approach.

ui.frontend/src/utils.js
arunattri26: Updated name as it was no linger being used as a registration function.
I will discuss with Rishi regarding separate util file.

girotraapankaj (approved):

PR #1567, opened 2025-03-27
feat(print-properties): Adding `dorContainer` in each component and `pageTemplate` in FormContainer and new print specific components

DhruvSharma-845:
## Description

- Adding `pageTemplate` in FormContainer to describe master page related properties for DoR.
- Adding `dorContainer` in each Component to describe print related properties for DoR.
- Print specific properties will be exported in CRISPR json only when present in JCR(i.e. in case of print only).
- Adding new components for print: PageSet, PageArea, Content, Medium and Proto. Keeping the implementations in a separate module.

## Motivation and Context

It will enable the support for print related properties in CRISPR json(mainly layout and style related properties).

## How Has This Been Tested?

UTs have been added.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have updated the documentation accordingly.
- [x] All unit tests pass on CircleCi.


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/Barcode.java
```diff
@@ -0,0 +1,21 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
rismehta, line 1: Please add these interfaces in a different package inside form

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -567,4 +585,20 @@ private List<String> getDisabledXFAScripts() {
         }
         return new ArrayList<>(disabledScripts);
     }
+
+    @JsonIgnore
+    public Map<String, Object> getDorContainer() {
+        if (resource != null) {
+            Resource dorContainerResource = resource.getChild("fd:dorContainer");
```
DhruvSharma-845, line 592: @rismehta : Here I am assuming `dorContainer` node will be named as `fd:dorContainer` in JCR, keeping it in consistency with `fd:rules` and `'fd:events`. I hope it is okay.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -402,9 +410,27 @@ public Map<String, Object> getDorProperties() {
...
+        Map<String, Object> pageTemplate = getPageTemplate();
+        if (pageTemplate != null) {
+            customDorProperties.put(DOR_PAGE_TEMPLATE, pageTemplate);
+        }
         return customDorProperties;
     }
 
+    private Map<String, Object> getPageTemplate() {
+        if (resource != null) {
+            Resource pageTemplateResource = resource.getChild("fd:pagetemplate");
```
DhruvSharma-845, line 422: `pageTemplate` node is assumed to be persisted with name `fd:pageTemplate`. I hope it is fine.

bundles/af-core/src/test/java/com/adobe/cq/forms/core/Utils.java
```diff
@@ -126,14 +128,14 @@ public static void testSchemaValidation(@NotNull Object model) {
         // create an instance of the JsonSchemaFactory using version flag
         JsonSchemaFactory schemaFactory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7);
         try {
-            InputStream schemaStream = Utils.class.getResourceAsStream("/schema/0.14.2/adaptive-form.schema.json");
+            InputStream schemaStream = Utils.class.getResourceAsStream("/schema/0.15.1/adaptive-form.schema.json");
```
DhruvSharma-845, line 131: @rismehta : Have added new schema files versioned 0.15.1. Also, have updated version number here in schema validation. Please let me know if anything else needs to be done w.r.t. new schema

bundles/af-core/src/test/resources/form/formcontainer/exporter-formContainer-cc-form-rest-submission.json
```diff
@@ -52,6 +52,6 @@
         "grammar": "json-formula-1.0.0",
         "version": "1.0.0"
     },
-    "adaptiveform": "0.14.2",
+    "adaptiveform": "0.15.1",
```
DhruvSharma-845, line 55: @rismehta :
Updating version number here in test baseline files is breaking a PR check. What is the right way to update this files?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/models/form/Barcode.java
DhruvSharma-845: Done

DhruvSharma-845:
@rismehta: In one of the architecture council meetings, we have decided to keep the value of `fragmentPath` in fragment object equals to the path till the actual fragment file node(for example; `/content/forms/af/testfragment`).
It means that FragmentImpl has to have the capability to resolve the appropriate channel of fragment depending on the channel of formcontainer being resolved.

Approach discussed in the meeting to add this capability has been attempted as follows:
- The new `PrintContainerImpl` will set the channel info equal to `print`. 
- This channel info will be passed down the hierarchy of objects through this code: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-374cea5eddcbd32818400130b79827999cbc1ab582332cc3c733df274d31bed6R147
- When the FragmentImpl received the channel info, it can decide the appropriate channel node based on it: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-4eaa72b52a20e6f3d8498af9d1b7d3a5209faa3cc188cb8c33b4ab5c426d629aR83

Can you please review it?

rismehta:
> @rismehta: In one of the architecture council meetings, we have decided to keep the value of `fragmentPath` in fragment object equals to the path till the actual fragment file node(for example; `/content/forms/af/testfragment`). It means that FragmentImpl has to have the capability to resolve the appropriate channel of fragment depending on the channel of formcontainer being resolved.
> 
> Approach discussed in the meeting to add this capability has been attempted as follows:
> 
> * The new `PrintContainerImpl` will set the channel info equal to `print`.
> * This channel info will be passed down the hierarchy of objects through this code: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-374cea5eddcbd32818400130b79827999cbc1ab582332cc3c733df274d31bed6R147
> * When the FragmentImpl received the channel info, it can decide the appropriate channel node based on it: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-4eaa72b52a20e6f3d8498af9d1b7d3a5209faa3cc188cb8c33b4ab5c426d629aR83
> 
> Can you please review it?

This could impact performance, as it applies to every resource in the hierarchy, in the web channel. While the impact would be noticeable mainly when the form size is huge (probably above 1000), it's still worth optimizing. Instead of setting the channel information at every level, can we fetch it just once from the root? That way, each fragment can simply query its  guide container for the channel information  (IIUC, setting request attribute won't work in this case, since request object isn't always available)

DhruvSharma-845:
> > @rismehta: In one of the architecture council meetings, we have decided to keep the value of `fragmentPath` in fragment object equals to the path till the actual fragment file node(for example; `/content/forms/af/testfragment`). It means that FragmentImpl has to have the capability to resolve the appropriate channel of fragment depending on the channel of formcontainer being resolved.
> > Approach discussed in the meeting to add this capability has been attempted as follows:
> > 
> > * The new `PrintContainerImpl` will set the channel info equal to `print`.
> > * This channel info will be passed down the hierarchy of objects through this code: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-374cea5eddcbd32818400130b79827999cbc1ab582332cc3c733df274d31bed6R147
> > * When the FragmentImpl received the channel info, it can decide the appropriate channel node based on it: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-4eaa72b52a20e6f3d8498af9d1b7d3a5209faa3cc188cb8c33b4ab5c426d629aR83
> > 
> > Can you please review it?
> 
> This could impact performance, as it applies to every resource in the hierarchy, in the web channel. While the impact would be noticeable mainly when the form size is huge (probably above 1000), it's still worth optimizing. Instead of setting the channel information at every level, can we fetch it just once from the root? That way, each fragment can simply query its guide container for the channel information (IIUC, setting request attribute won't work in this case, since request object isn't always available)

Yes, request attribute cannot be relied upon here because it will not be available in case we are "adapting to" from a resource.
I agree that it will impact every adaptation in the hierarchy. So, as you suggested, what I have tried is
- FragmentImpl will search for the guideContainer up in the hierarchy. 
- It will determine the channel of the fragment based on the resourceType of the guideContainer.
This way, only the fragment resolution will have the extra overhead.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -567,4 +585,26 @@ private List<String> getDisabledXFAScripts() {
...
+     * 
+     * @return Map of dor container properties if `fd:dorContainer` node is present otherwise returns null
+     */
+    @JsonIgnore
+    @Override
+    public Map<String, Object> getDorContainer() {
+        if (resource != null) {
+            Resource dorContainerResource = resource.getChild("fd:dorContainer");
+            if (dorContainerResource != null) {
+                DorContainer dorContainer = dorContainerResource.adaptTo(DorContainer.class);
```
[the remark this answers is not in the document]
sakshi-arora1: Do we have a single interface for all components? (this seems to be the case, seeing the implementation), how large is it? if its very small, and will not be required to change frequently, then we can have it here, this will lead to removing of dependency of core components on newly introduced print artifact

sakshi-arora1 (changes requested):

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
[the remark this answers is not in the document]
DhruvSharma-845: > this is confusing -> DorContainer for each component, nomenclature needs to be improved, I believe you mean DorComponent and not DorContainer

The name `DorContainer` was decided during the spec review and it has been approved and integrated in schema as well:
- https://git.corp.adobe.com/livecycle/af2-docs/issues/218
- https://git.corp.adobe.com/livecycle/af2-docs/blob/gh-pages/schema/print/adaptive-form-dor-container-properties.schema.yaml


> Do we have a single interface for all components? (this seems to be the case, seeing the implementation), how large is it? if its very small, and will not be required to change frequently, then we can have it here, this will lead to removing of dependency of core components on newly introduced print artifact

Yes...we have a single interface for all components because they have a large number of common properties and almost half of them are non-primitive(objects of other classes). Since there were so many classes involved, that's why, @rismehta suggested moving these interfaces to a different artifact.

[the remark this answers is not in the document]
sakshi-arora1: ok, so basically it encapsulates all other classes. Then we might have to live with print dependency. Just make sure that only interfaces are defined in the dependant print artifact. The implementations of these apis are in a different artifact (bundle), this will ensure keeping the dependency light and will be less prone to breaking core components.

sakshi-arora1 (approved):
approving it basis, its already approved in spec review.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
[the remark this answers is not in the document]
DhruvSharma-845: To keep it lean, we have even segregated the interfaces into two bundles.
`print-dor-api` just contains the interfaces that are must for Core Components to work.

vdua:
I have one doubt regarding the dorContainer. I thought the dorContainer properties will come with different View (not author/publish) ? I do not see that change here ?

DhruvSharma-845:
> I have one doubt regarding the dorContainer. I thought the dorContainer properties will come with different View (not author/publish) ? I do not see that change here ?

I have added `@JsonView` annotation above getDorContainer method. Please refer here: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-83ed79b144226c70d6b93ceaa9b25d5d243edacab1730e932f106c5e02a2df0cR84

I took the inspiration from getDorProperties method: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-83ed79b144226c70d6b93ceaa9b25d5d243edacab1730e932f106c5e02a2df0cL73

vdua:
> > I have one doubt regarding the dorContainer. I thought the dorContainer properties will come with different View (not author/publish) ? I do not see that change here ?
> 
> I have added `@JsonView` annotation on getDorContainer method. Please refer here: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-83ed79b144226c70d6b93ceaa9b25d5d243edacab1730e932f106c5e02a2df0cR84
> 
> I took the inspiration from getDorProperties method: https://github.com/adobe/aem-core-forms-components/pull/1567/files#diff-83ed79b144226c70d6b93ceaa9b25d5d243edacab1730e932f106c5e02a2df0cL73

Sorry. Missed that. Looks okay to me.

vdua (approved):
LGTM. Ensure 
* the tests pass before merging
* use squash and merge rather than adding these 571 commits to the repo.

PR #1572, opened 2025-04-09
fix: Add null check for span element in radio button component

im-shiv:


ui.frontend/src/view/FormOptionFieldBase.js
```diff
@@ -122,7 +122,9 @@ class FormOptionFieldBase extends FormFieldBase {
         } else {
             [...this.getOptions()].forEach((option, index) => {
                 let span = option.querySelector('span');
-                span.innerHTML = window.DOMPurify ?  window.DOMPurify.sanitize(newEnumNames[index]) : newEnumNames[index];
+                if(span) {
```
rismehta, line 125: Please add NPE check above as well. Also, let's write a test case for this similar to the use-case

rismehta (changes requested):
check comments

PR #1610, opened 2025-06-10
Fixing positional enter in text area.

sakshi-arora1:
## Types of changes

## Checklist:


devgurjar (approved):

rismehta:
Why was this merged without test case ?

rismehta:
> > Why was this merged without test case ?
> 
> @rismehta we already have a test case for newline charachter https://github.com/adobe/aem-core-forms-components/blob/dev/ui.tests/test-module/specs/textinput/textinput.runtime.cy.js#L150
> 
> and it seems we can't go to a cursor location to type using cypress. I searched that, and all the solutions were to put the whole value including \n. The fix here is particularly to type \n in between.

If we already have a test case, then why do we need this fix ?

PR #1700, opened 2025-09-03
Adding compilation option for all client libraries.

sakshi-arora1:
## Types of changes

## Checklist:

### DIFF
--- a/it/apps/src/main/content/jcr_root/apps/forms-core-components-it/clientlibs/clientlib-it-base/.content.xml
+++ b/it/apps/src/main/content/jcr_root/apps/forms-core-components-it/clientlibs/clientlib-it-base/.content.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <jcr:root xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:cq="http://www.day.com/jcr/cq/1.0"
     jcr:primaryType="cq:ClientLibraryFolder"
+    jsProcessor="[default:none,min:gcc;languageIn=ECMASCRIPT_2021;languageOut=ECMASCRIPT_2021]"
     allowProxy="{Boolean}true"
     categories="[forms-it.base]"/>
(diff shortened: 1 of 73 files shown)


rombert (commented):
@sakshi-arora1 - I am not sure this does what you want it to.

When precompiling client libraries during a full stack pipeline execution we use closure compiler v20210106 .

Looking at the source code I don't see ECMASCRIPT_2021 as a valid option for languageIn - https://github.com/google/closure-compiler/tree/v20211006 .

> Sets the language spec to which input sources should conform. Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, ECMASCRIPT_2015, ECMASCRIPT_2016, ECMASCRIPT_2017, ECMASCRIPT_2018, ECMASCRIPT_2019, STABLE, ECMASCRIPT_NEXT

rismehta (changes requested):
We should just add it for editor specific client libraries, since runtime client libraries are always included via runtime.all which already has this.

sakshi-arora1:
> We should just add it for editor specific client libraries, since runtime client libraries are always included via runtime.all which already has this.

But, its upto customer they way they wish to include these individual client libs. Do you see any harm in putting this on each clientlib?

rismehta:
> > We should just add it for editor specific client libraries, since runtime client libraries are always included via runtime.all which already has this.
> 
> But, its upto customer they way they wish to include these individual client libs. Do you see any harm in putting this on each clientlib?

Let's just fix the original issue for now to keep the impact minimal

sakshi-arora1:
> > > We should just add it for editor specific client libraries, since runtime client libraries are always included via runtime.all which already has this.
> > 
> > 
> > But, its upto customer they way they wish to include these individual client libs. Do you see any harm in putting this on each clientlib?
> 
> Let's just fix the original issue for now to keep the impact minimal

But how will this impact anything?

sakshi-arora1:
> @sakshi-arora1 - I am not sure this does what you want it to.
> 
> In the FACT tool we embed closure compiler v20210106 - https://git.corp.adobe.com/Granite/com.adobe.granite.fact/blob/ad2ab8bf993d0c92ded6b31be81aeded997b921a/tool/pom.xml#L773-L777 .
> 
> Looking at the source code I don't see ECMASCRIPT_2021 as a valid option for languageIn - https://github.com/google/closure-compiler/tree/v20211006 .
> 
> > Sets the language spec to which input sources should conform. Options: ECMASCRIPT3, ECMASCRIPT5, ECMASCRIPT5_STRICT, ECMASCRIPT_2015, ECMASCRIPT_2016, ECMASCRIPT_2017, ECMASCRIPT_2018, ECMASCRIPT_2019, STABLE, ECMASCRIPT_NEXT

Hi @rombert , that is correct, I looked up and it seems ECMASCRIPT_2021 was introduced in version v20211107 [1].
We have a lot of existing clientlibraries which are actually already pointing to ECMASCRIPT_2021, and when I run them through the com.adobe.granite.clientlibs.compiler it doesn't error out for them. I will set ECMASCRIPT_NEXT and check.


[1] https://github.com/google/closure-compiler/wiki/Flags-and-Options/c1546742e8d7be056afa3621b15b10bf43206547

Issue #1215, opened 2024-05-02
Form allows duplicate submits


mnahmias:
### Expected Behaviour
Form should have some form of disabled or in progress state while a submit is in progress to prevent multiple clicks of the submit button triggering multiple form submissions.

### Actual Behaviour
The form doesn't change, allowing the user to repeatedly click the submit button.

#### AEM Version (mention the exact version in case of cloud SDK)
2023.12.14697.20231215T125030Z-231200

#### AEM Forms Version
2023.10.25.00-231200

rismehta:
@mnahmias This feature isn't available out-of-the-box yet, but you can modify the submission process to include this logic.

rismehta:
> Is anyone working on this? It is a very basic feature that Adaptive Forms with Foundations components had from day one. It should be available for Core Components without the necessity to customize the submission process.

@martin-grimm Agree. I have asked the team to expedite the resolution for this.

Nadine-Herczeg:
Is there any information on the status for the issue here yet?

armaang1729:
This issue has been resolved and the latest version having the fix would be available in the november release. 

Thanks.

PR #1713, opened 2025-09-15
Feat: added the loader to avoid multiple submits and tests

armaang1729:
# Added loader to prevent multiple form submits 

## Description

This PR addresses [Issue #1215](https://github.com/adobe/aem-core-forms-components/issues/1215), which reported that clicking the **Submit** button multiple times triggered multiple form submissions, resulting in duplicate entries.

To resolve this, a loader has been introduced that appears on form submission. While the loader is active, the form is disabled, preventing duplicate submits until the process is complete.

## Related Issue

- Fixes [Issue #1215](https://github.com/adobe/aem-core-forms-components/issues/1215)

## Motivation and Context

Multiple submissions from repeated clicks on the **Submit** button were leading to duplicate form entries, which negatively impacted both user experience and backend data integrity.  
By introducing a loader that disables the form during submission, this fix ensures that only one submission is processed at a time.

## How Has This Been Tested?

- Added Cypress tests to verify that the loader class appears when the form is submitted.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] All unit tests pass on CircleCi.

### DIFF
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/button/v1/button/clientlibs/site/js/buttonview.js
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/button/v1/button/clientlibs/site/js/buttonview.js
@@ -69,14 +69,14 @@
         _showLoader() {
             const container = this._getContainer();
             if (container) {
-                container.classList.add('cmp-adaptiveform-container--loading');
+                container.classList.add('cmp-adaptiveform-container--submitting');
             }
         }
 
         _hideLoader() {
             const container = this._getContainer();
             if (container) {
-                container.classList.remove('cmp-adaptiveform-container--loading');
+                container.classList.remove('cmp-adaptiveform-container--submitting');
             }
         }
 
--- a/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/button/v1/button/clientlibs/site/js/buttonview.js
+++ b/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/button/v1/button/clientlibs/site/js/buttonview.js
@@ -13,7 +13,7 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/
-(function() {
+  (function() {
 
     "use strict";
     class Button extends FormView.FormFieldBase {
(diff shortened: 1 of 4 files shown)


ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/button/v1/button/clientlibs/site/js/buttonview.js
```diff
@@ -36,6 +36,59 @@
...
+                // End loading on submit outcomes
+                formModel.subscribe(() => {
+                    this._hideLoader();
+                }, 'submitSuccess');
+
+                formModel.subscribe(() => {
+                    this._hideLoader();
+                }, 'submitError');
+
+                this._formEventHandlersInstalled = true;
```
sakshi-arora1, line 87: this will not work when form is submitted again post validation failure

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/button/v1/button/clientlibs/site/js/buttonview.js
```diff
@@ -36,6 +36,50 @@
...
+        _container = null;
+
+        _getContainer() {
+            if (this._container) return this._container;
+            const container = this.formContainer?.getFormElement?.();
+            if (container) this._container = container;
+            return this._container;
+        }
+
+        _showLoader() {
```
rismehta, line 69: Should we also disable the form ? Problem with this approach is it won't work on EDS since the changes are view layer specific

Why can't we just disable the form on submit and enable it when validation is complete.

rismehta: > we should not enable disable form, as customers can have rules which actually disables form by some rule, and we can accidentally enable it back

This can be addressed by maintaining a state. I’d prefer not to make assumptions here, since the behavior is customer-driven. 

@armaang1729 Can we review how it was implemented in v1 and reapply the same approach, given that this was specifically requested by a customer?

rismehta (changes requested):
check comments

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/button/v1/button/clientlibs/site/js/buttonview.js
armaang1729: Checked in V1 also the form is not disabled there.

PR #1714, opened 2025-09-16
FORMS-19439 adding support for display format in text input

rismehta:
## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
@@ -30,7 +30,6 @@ import java.util.stream.StreamSupport;
 
 import javax.servlet.Servlet;
 import javax.servlet.ServletException;
-import javax.servlet.http.HttpServletRequest;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.jackrabbit.JcrConstants;
@@ -143,7 +142,7 @@ public class FormMetaDataDataSourceServlet extends AbstractDataSourceServlet {
             FormMetaDataType type = FormMetaDataType.fromString(getParameter(config, TYPE, request, null));
             String dataModel = getParameter(config, DATA_MODEL, request, "");
             actionTypeDataSource = new SimpleDataSource(getDataSourceResources(
-                request, request.getResourceResolver(), type, dataModel).iterator());
+                request, request.getResourceResolver(), type, dataModel, config).iterator());
         }
         request.setAttribute(DataSource.class.getName(), actionTypeDataSource);
     }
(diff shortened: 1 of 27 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
```diff
@@ -200,7 +199,15 @@ private List<Resource> getDataSourceResources(HttpServletRequest request, Resour
                             }
                         }
                     }
-                    resources.add(getResourceForDropdownDisplay(resourceResolver, "Custom", "custom"));
+                    // Only add "Custom" option for non-text-input components
```
sakshi-arora1, line 202: Why? we can have a custom format for text input as well?

rismehta: I don’t have a use case for this. “Custom” implies that authors must manually provide input. In this scenario, we were adding a displayValueExpression for all dropdown values, which internally calls an OOTB function. Since this isn’t possible for custom values, it has been removed.

rismehta:
> Additionally formats will be needed for textinput:
> 
> 1. cq-guide https://git.corp.adobe.com/livecycle/cq-guides/tree/master/content/src/main/resources/libs/settings/wcm/policies/core/fd/components/form
> 2. archetype https://github.com/adobe/aem-project-archetype

Yes, it will be added here later

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/af-commons/v1/clientlibs/editor/utils/utils.js
```diff
@@ -177,17 +177,21 @@
...
+                                displayFormatParentDiv.setAttribute("hidden", true);
+                                break;
+                            default           :
+                                displayFormatParentDiv.removeAttribute("hidden");
+                        }
                     }
                 }
-                if(displayPatternSelectedValue!="custom") {
+                // Only update format component value if it exists and is not hidden
+                if (formatComponent && !formatComponent.closest("div").hasAttribute("hidden") && displayPatternSelectedValue!="custom") {
```
muskgupta: @rismehta 
The check that has been introduced here is causing a bug in this scenario - dropdown is not visible, only textfield to enter the pattern is visible - and user enters some value in it.
This case makes (formatComponent && !formatComponent.closest("div").hasAttribute("hidden")) condition true and causes formatComponent to update its value to "" (as patternComponent is hidden and value is empty) which should not be the case.

cc: @sakshi-arora1

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/servlets/FormMetaDataDataSourceServlet.java
muskgupta: @rismehta Removing Custom is causing issues as the field where the pattern is reflected basis dropdown selection is editable and the user can add custom values there - which should redirect to Custom in dropdown. Since custom is not available right now it is causing inconsistency and redirecting to Select.
JIRA https://jira.corp.adobe.com/browse/FORMS-23596

rismehta: Please add relevant test case for the same, the issue from description is not yet clear.

PR #1752, opened 2025-11-19
Unload

rismehta:
## Types of changes

## Checklist:

### DIFF
--- a/ui.frontend/src/GuideBridge.js
+++ b/ui.frontend/src/GuideBridge.js
@@ -494,6 +494,11 @@ class GuideBridge {
             }
         }
 
+        // Clear XFA-specific state if XFA forms are loaded
+        if (window.formBridge && typeof window.formBridge.destroyForm === 'function') {
+            window.formBridge.destroyForm();
+        }
+
         // Clean up global state when this is the last/only form
         // NOTE: Global widgets, caches, and utilities don't have form-specific identifiers,
         // so we can only safely clear them if this is the only/last form on the page.
@@ -522,12 +527,9 @@ class GuideBridge {
             // Clear Utils static state (contextPath, fieldCreatorSets, fieldCreatorOrder)
             Utils._clearState();
 
-            // Clear XFA-specific state if XFA forms are loaded
+            // Clear XFA utility (global XFA logger/utility)
             if (window.xfalib && window.xfalib.ut) {
                 window.xfalib.ut = null;
-            }
-            if (window.formBridge && typeof window.formBridge.destroyForm === 'function') {
-                window.formBridge.destroyForm();
             }
         }
 


ui.frontend/src/GuideBridge.js
```diff
@@ -451,6 +451,84 @@ class GuideBridge {
...
+        delete this.#formContainerViewMap[pathToUnload];
+
+        // Reset formContainerPath if it matches the one being unloaded
+        if (this.#formContainerPath === pathToUnload) {
+            this.#formContainerPath = "";
+        }
+
+        // Note: FunctionRuntime.customFunctions is intentionally NOT cleared
+        // as it's a global registry shared across all forms, matching classic behavior
+    }
```
vdua: We also have mutationObservers ? We should clear those as well if not doing them

ui.frontend/src/GuideBridge.js
```diff
@@ -451,6 +451,84 @@ class GuideBridge {
...
+            container = formContainerView.getFormElement();
+        }
+
+        // Clean up widget elements appended to document.body
+        // These widgets are created by captcha components and datepicker that append to body
+        const widgetSelectors = [
+            '.cmp-adaptiveform-recaptcha__widget',
+            '.cmp-adaptiveform-turnstile__widget',
+            '.cmp-adaptiveform-hcaptcha__widget',
+            '.datePickerTarget'
```
vdua: This points to the fact that we need to have a better mechanism for widget selectors.

ui.frontend/src/GuideBridge.js
```diff
@@ -451,6 +451,84 @@ class GuideBridge {
...
+        ];
+        
+        widgetSelectors.forEach(selector => {
+            const widgets = document.querySelectorAll(selector);
```
vdua: we should be selecting these inside the particular form which is being removed.

ui.frontend/src/GuideBridge.js
rismehta: Observer will eventually be garbage collected, but it's not immediate and could cause temporary memory overhead. Let me fix this as well

ui.frontend/src/GuideBridge.js
rismehta: yes, this cannot be changed at this time, most probably this selector came from v1

ui.frontend/src/GuideBridge.js
rismehta: These widgets are shared/global resources, and removing ALL of them will break other forms, its just a temporary state across all forms.

I can check if there is only one form left, when unload is called, we can clear both cache and this storage so that it is immediately garbage collected

PR #1821, opened 2026-02-13
FORMS-24087 Enabling FT to merge fragment container rules/events into…

rismehta:
… fragment placeholder

## Types of changes

## Checklist:


sakshi-arora1 (commented):
Why do we have toggle.api jar checked in?

bundles/af-core/lib/README.md
```diff
@@ -0,0 +1,9 @@
+# Runtime-only dependencies
+
+This folder is used for JARs that are not in a Maven repository but are available in AEM at runtime.
+
+## com.adobe.granite.toggle.api
+
+Place `com.adobe.granite.toggle.api-1.2.8.jar` here for compilation. The API is provided by AEM at runtime.
+
+You can obtain the JAR from your AEM SDK installation or from an AEM runtime (e.g. from the relevant bundle in the running instance).
```
sakshi-arora1: can we fetch it from internal repo rather, we already are accessing that for quickstart loads?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/FragmentImpl.java
```diff
@@ -83,6 +86,24 @@ public class FragmentImpl extends PanelImpl implements Fragment {
...
+    /**
+     * This bundle's context, provided by {@link BundleContextServiceRegistration}. Required for
+     * feature-toggle lookups (e.g. {@link ComponentUtils#isToggleEnabled}); without that component
+     * this would be null because BundleContext is not normally published as an OSGi service.
+     */
+    @OSGiService(
+        filter = "(" + BundleContextServiceRegistration.SERVICE_PROPERTY_PROVIDER + "="
+            + BundleContextServiceRegistration.SERVICE_PROPERTY_PROVIDER_VALUE + ")",
+        injectionStrategy = InjectionStrategy.OPTIONAL)
+    private BundleContext bundleContext;
```
sakshi-arora1: why can't we use following?
com.adobe.granite.toggle.api.ToggleRouter

@Reference(policyOption = ReferencePolicyOption.GREEDY)
    protected ToggleRouter toggleRouter;

sakshi-arora1: we can pass the toggle router further, in case you want it as a utility function

bundles/af-core/lib/README.md
rismehta: Its not part of public maven repo, but available at runtime

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/FragmentImpl.java
rismehta: It isn't always available (for example) on local SDK by default its not present. The code was failing in CI

bundles/af-core/lib/README.md
sakshi-arora1: its part of internal artifactory

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/FragmentImpl.java
sakshi-arora1: But if sdk is not exposing com.adobe.granite.toggle.api, then it means it can not be used. We are just force fitting via a workaround, we might need to get this added to sdk

rismehta: > But if sdk is not exposing com.adobe.granite.toggle.api, then it means it can not be used. We are just force fitting via a workaround, we might need to get this added to sdk

Ideally, no — since it’s not intended for end users. In this case, the core component exists in the public repository, but it is actually part of the add-on only. Therefore, we cannot expose it in the SDK. It’s more of an edge case. This was also one of the main reason we did not rely on FT yet in this repo, but given the regression we are having, we need a way out.

bundles/af-core/lib/README.md
rismehta: > its part of internal artifactory

yes, but available at runtime on cloud service.

PR #1879, opened 2026-05-11
Added Table component support in Core Components 

armaang1729:
## Types of changes

## Checklist:


examples/ui.apps/src/main/content/jcr_root/apps/forms-components-examples/components/form/table/_cq_template.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  ~ Copyright 2024 Adobe
```
iamsudhanshu: 2026

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/replacehook.js
```diff
@@ -30,8 +30,10 @@
         'checkbox-group': fieldTypes.SELECT,
         'checkbox': fieldTypes.CHECKBOX,
         'date-input': fieldTypes.TEXT,
+        'datetime-input': fieldTypes.TEXT,
```
armaang1729, line 33: NOTE :

Previously, the replace dialog restricted targets to the same `typeMap` family (e.g. `text-input` could only be replaced with TEXT-family components). This was too limiting inside table row cells where any policy-allowed component should be swappable.

Changes:
- `isUnderCoreTableRow`: detects when the editable is inside a table row/header and skips the type-family check entirely for those cells.
- `buildAllowedTemplatePathSet`: still enforces the parent container's policy, so only components the row actually allows appear in the list.
- `getTemplateFieldType` + null guard: fixes a crash when a component's `.json` omits `fieldType` — now falls back to `model.json`.
- Added `datetime-input` and `multiline-input` to `typeMap` so they match correctly in standard (non-table) replace flows.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/tableroweditorhook.js
```diff
@@ -0,0 +1,1108 @@
+/*******************************************************************************
```
armaang1729, line 1: NOTE :

Author-side editor hook for the core table component. Powers all structural editing actions on a table in the AEM page editor — add/delete/reorder rows, add/delete columns, and merge/split cells — by making Sling POST requests to manipulate JCR nodes directly and refreshing the table editable. Also patches the AEM overlay renderer to correctly position edit overlays on `<tr>` elements, which can get misplaced due to browser foster-parenting of table markup.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/toolbaractionhook.js
```diff
@@ -17,10 +17,36 @@
     "use strict";
     let superSitesEditorAppendButton = ns.edit.Toolbar.prototype.appendButton;
 
+    /**
+     * Table row cells must not expose the standard Delete action (row/column flows handle structure).
+     * @param {Granite.author.Editable} editable
```
armaang1729, line 22: NOTE :

Table row cells must not expose the standard Delete action like normal components. The rows must be deleted through the delete row option added in the edit config of the table row.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/js/tableview.js
```diff
@@ -0,0 +1,233 @@
+/*******************************************************************************
```
armaang1729, line 1: NOTE :

**tableview.js — purpose**

This file is the runtime view for the Adaptive Form Table component. It has three responsibilities:

1. **Row insertion** () — inserts a cloned `<tr>` into the correct `<tbody>` position using the row model's index. Handles three cases: empty tbody (append), index 0 (prepend before the first sibling), and mid-list (insert after the preceding sibling resolved by model index).

2. **Hook attribute patching** (`#syncTableRowHooks`) — `FormView.Utils.updateId` patches element `id` attributes on cloned nodes but does not touch `data-cmp-hook-*` attributes. Without this step, cloned add/remove buttons retain the template row's stale ID, causing every dynamically-added row to dispatch the wrong model index. This method re-stamps `data-cmp-hook-add-instance` and `data-cmp-hook-remove-instance` with the newly assigned model ID immediately after insertion.

3. **min/maxOccur enforcement** (`#syncTableRowRepeatableControls`) — after every addition or removal, walks every live row and sets `data-cmp-visible` on its add and remove buttons according to the instance manager's current item count vs. `minOccur`/`maxOccur`. A `maxOccur` of `-1` means unbounded; `minOccur` of `-1` means no floor.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/text/v1/text/_cq_editConfig.xml
```diff
@@ -150,5 +150,19 @@
                 handler="CQ.FormsCoreComponents.editorhooks.viewQualifiedName"
                 icon="viewSOMExpression"
                 text="View Qualified Name"/>
+        <addcolumn
```
armaang1729, line 153: NOTE :

Added addColumn and deleteColumn option in the text editconfig as the option to add column and delete will be displayed when the column cell is clicked which will be a text component so added a condition `isCoreTableHeaderCell(editable)` to display this option.

ui.frontend/src/view/FormPanel.js
```diff
@@ -36,6 +36,24 @@ class FormPanel extends FormFieldBase {
         this.children = [];
     }
 
+    /**
+     * DOM container where repeatable sibling instances live (InstanceManager parentElement,
+     * e.g. for getElementAt). Override in layout-specific panels (table row: tbody / table body div).
+     * @returns {HTMLElement|undefined}
+     */
+    getRepeatableInstancesContainerElement() {
+        return this.element.parentElement?.parentElement;
```
armaang1729, line 45: New virtual hook that returns the DOM container holding all sibling repeat instances. The default (`parentElement?.parentElement`) covers the standard Sling-wrapped panel structure. Subclasses like `TableRow` override this to return `<tbody>` instead, since a `<tr>` sits at a different nesting depth. `instantiateInstanceManager()` reads this so `InstanceManager` knows where to append/insert cloned instances.

ui.frontend/src/view/FormPanel.js
```diff
@@ -36,6 +36,24 @@ class FormPanel extends FormFieldBase {
...
+    }
+
+    /**
+     * DOM node cloned as the repeat template and removed with the instance.
+     * Default: parent of the field root. Override when the repeatable unit differs (e.g. Sling wrapper vs tr).
+     * @returns {HTMLElement|undefined}
+     */
+    getRepeatableDomWrapper() {
+        return this.element.parentElement;
```
armaang1729, line 54: Virtual hook that returns the node to clone as the repeat template and to remove when an instance is deleted. Default is `parentElement` (the Sling wrapper div). `TableRow` overrides this to return the `<tr>` itself. Without this hook, `updateTemplate` and `handleRemoval` in `InstanceManager` would always clone/remove the wrong element for table rows.

ui.frontend/src/view/InstanceManager.js
```diff
@@ -135,8 +133,11 @@ class InstanceManager {
          this.parentElement.insertBefore(markerElement, repeatableElement);
          this.markerElement = markerElement;
          **/
+        const wrapper = (typeof childView.getRepeatableDomWrapper === "function")
+            ? childView.getRepeatableDomWrapper()
```
armaang1729, line 137: Uses the view's declared wrapper node (e.g. `<tr>` for a table row) instead of always `parentElement`. Cloning the wrong node would produce a template with a mismatched tag that breaks table structure when inserted into `<tbody>`.

ui.frontend/src/view/InstanceManager.js
```diff
@@ -173,7 +174,11 @@ class InstanceManager {
             console.error('Panel needs to have templateHTML to support repeatability.');
             return;
         }
-        return this.handleAddition(addedModel, beforeElement);
+        const addedHtmlElement = this.handleAddition(addedModel, beforeElement);
+        if (addedHtmlElement) {
```
armaang1729, line 178: `createFieldsForAddedElement` was previously called in two separate branches of `#syncViewModel`. Consolidating it here means any path through `#addChildInstance` automatically gets field initialization — no risk of a future caller forgetting it.

ui.frontend/src/view/InstanceManager.js
```diff
@@ -182,15 +187,27 @@ class InstanceManager {
...
     #removeChildInstance(removedModel) {
-        const removedIndex = removedModel.index;
-        let removedChildView = this.children[removedIndex];
-        if (removedIndex == -1) {
-            //That is, model was removed by prefill, and instance manager was synced with child already removed
+        if (!removedModel || !removedModel.id) {
+            return;
+        }
+        let removedChildView = this.children.find((cv) => cv && cv.getId() === removedModel.id);
+        if (!removedChildView && typeof removedModel.index === "number" && removedModel.index >= 0
```
armaang1729, line 194: Changed from index-based to id-based lookup because `model.index` can be stale after a prior remove shifts the `children` array. The index fallback is kept only for cases where the view hasn't been registered by id yet (e.g. prefill-removed instances). `this.children.indexOf(removedChildView)` at the end re-derives the live array position rather than trusting the model value.

ui.frontend/src/view/InstanceManager.js
```diff
@@ -314,8 +331,15 @@ class InstanceManager {
             //give parentView chance to adjust to removed instance
             this.repeatableParentView.handleChildRemoval(removedInstanceView);
         }
-        //removing just the parent view HTML child instance, to avoid repainting of UI for removal of each child HTML
-        removedInstanceView.element.parentElement.remove();
+        let toRemove;
+        if (typeof removedInstanceView.getRepeatableDomWrapper === "function") {
```
armaang1729, line 335: Delegates to `getRepeatableDomWrapper` so the correct node (e.g. the whole `<tr>`) is removed. Without this, removing a table row would call `.remove()` on a `<div>` wrapper that doesn't exist in the DOM, silently doing nothing.

ui.frontend/src/utils.js
```diff
@@ -144,13 +144,17 @@ class Utils {
...
      * @param {Element} addedElement - The added element.
      * @param {module:FormView~FormContainer} formContainer - The form container.
      */
     static createFieldsForAddedElement(addedElement, formContainer) {
         Object.values(Utils.#fieldCreatorOrder).forEach(function (fieldSelector) {
             let fieldCreatorSet = Utils.#fieldCreatorSets[fieldSelector];
-            const fieldElements = addedElement.querySelectorAll(fieldCreatorSet['fieldSelector']);
+            const fieldElements = Array.from(addedElement.querySelectorAll(fieldCreatorSet['fieldSelector']));
+            if (addedElement.matches && addedElement.matches(fieldCreatorSet['fieldSelector'])) {
+                fieldElements.unshift(addedElement);
```
armaang1729, line 156: `querySelectorAll` only searches descendants — it never matches the element itself. When `addedElement` is a `<tr data-cmp-is="adaptiveFormTableRow">`, the existing call returned nothing and the row's view was never created. The `matches` check at the root fixes this for any component whose outermost element carries `data-cmp-is`.

ui.frontend/src/utils.js
```diff
@@ -203,10 +207,13 @@ class Utils {
      * @param {module:FormView~FormContainer} formContainer - The form container.
      */
     static #removeChildReferences(fieldView, formContainer) {
+        if (!fieldView) {
+            return;
```
armaang1729, line 211: Null guard added so a missing `fieldView` exits early instead of throwing on `fieldView.children`.

ui.frontend/src/utils.js
```diff
@@ -203,10 +207,13 @@ class Utils {
...
+        }
         let childViewList = fieldView.children;
         if (childViewList) {
             for (let index = 0; index < childViewList.length; index++) {
-                Utils.#removeChildReferences(childViewList[index]);
+                Utils.#removeChildReferences(childViewList[index], formContainer);
             }
```
armaang1729, line 217: The recursive call was previously missing `formContainer`, so it received `undefined` and silently skipped `#removeFieldId` for all grandchild views. Deep nested panels (e.g. a panel inside a repeatable table row) would leak their field references in the form container's field map.

ui.frontend/src/utils.js
```diff
@@ -222,22 +229,30 @@ class Utils {
...
     /**
      * Update the id inside the given html element.
+     * Also checks if the root element itself has the id (querySelectorAll only matches descendants).
      * @param {Element} htmlElement - The HTML element.
      * @param {string} oldId - The old ID.
      * @param {string} newId - The new ID.
      */
     static updateId(htmlElement, oldId, newId) {
+        if (htmlElement.id === oldId) {
+            htmlElement.id = newId;
```
armaang1729, line 253: Same root vs. descendants issue as `createFieldsForAddedElement`. When `htmlElement` is the cloned `<tr>` and its `id` needs to be swapped, `querySelectorAll('#oldId')` finds nothing because it only searches children. The early return on a direct id match handles this before falling through to the descendant search.

armaang1729 (commented):
Inline comments explaining the key changes in FormPanel.js, InstanceManager.js, and utils.js.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/tablerow/v1/tablerow/clientlibs/site/js/tablerowview.js
```diff
@@ -0,0 +1,117 @@
...
+            this.children = [];
+        }
+
+        /**
+         * For table rows in a single tbody, the row element is the repeatable unit.
+         * In edit mode with div wrappers, return the Sling wrapper div.
+         * @returns {HTMLElement}
+         */
+        getRepeatableDomWrapper() {
+            const parent = this.element.parentElement;
```
armaang1729, line 39: NOTE :

This is the consumer of the virtual hook defined in `FormPanel`. The check for `bemBlock` on `parentElement` handles the edit-mode case where Sling wraps the `<tr>` in a `<div class="cmp-adaptiveform-tablerow">` for the authoring overlay. In publish mode the `<tr>` has no such wrapper, so `this.element` (the `<tr>` itself) is returned directly. `InstanceManager` uses this for both cloning the template and tearing down removed rows.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/tablerow/v1/tablerow/clientlibs/site/js/tablerowview.js
```diff
@@ -0,0 +1,117 @@
...
+            }
+            return this.element;
+        }
+
+        /**
+         * Container where sibling rows are inserted — the {@code <tbody>} or div.table__body.
+         * @returns {HTMLElement}
+         */
+        getRepeatableInstancesContainerElement() {
+            const tableBody = this.element.closest(".cmp-adaptiveform-table__body") ||
```
armaang1729, line 51: NOTE : 

Returns `<tbody>` as the container so `InstanceManager.handleAddition` inserts cloned rows into the right parent. Using `closest` handles both the semantic `<tbody class="cmp-adaptiveform-table__body">` and any fallback structure, with `parentElement` as a last resort to stay consistent with the base class default.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/tablerow/v1/tablerow/clientlibs/site/js/tablerowview.js
```diff
@@ -0,0 +1,117 @@
...
+            return this.getId();
+        }
+
+        /**
+         * Override updateLabel to handle the table row's specific HTML structure.
+         * Table rows don't have labels in the traditional sense.
+         * @param {Object} label - The label state object.
+         */
+        updateLabel(label) {
+            // Table rows don't have visible labels, so this is a no-op
```
armaang1729, line 99: NOTE : 

No-op intentionally — table rows have no visible label of their own (column headers are owned by `tableheader`). Without this override, `applyState` would call the base class `updateLabel` which would try to update a non-existent label DOM element.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/tablerow/v1/tablerow/clientlibs/site/js/tablerowview.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        /**
+         * Override applyState for table row's specific structure.
+         * @param {Object} state - The state to be applied.
+         */
+        applyState(state) {
+            this.updateVisible(state.visible);
```
armaang1729, line 107: NOTE : 

Matches the `FormPanel.applyState` signature but explicitly skips anything table-row-irrelevant (e.g. `updateLabel` is a no-op here). Calling `initializeHelpContent` is kept because a row's visibility/enabled state still drives `data-cmp-visible` / `data-cmp-enabled` attributes that the rules engine reads.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/tablerow/v1/tablerow/clientlibs/site/js/tablerowview.js
```diff
@@ -0,0 +1,117 @@
...
+            this.updateEnabled(state.enabled);
+            this.initializeHelpContent(state);
+            this.updateLabel(state.label);
+        }
+    }
+
+    FormView.Utils.setupField(({element, formContainer}) => {
+        return new TableRow({element, formContainer})
```
armaang1729, line 115: NOTE : 

Registers `TableRow` as the view for all elements matching `[data-cmp-is="adaptiveFormTableRow"]`. This is the standard `FormView.Utils.setupField` pattern — it both creates the view for elements already in the DOM and wires up the mutation observer so dynamically added rows (repeat add) are initialized automatically.

ui.tests/test-module/specs/repeatable/repeatable.updatelabel.cy.js
```diff
@@ -54,10 +54,27 @@ describe("Form Runtime with Panel Container", () => {
                 checkLabelText(textinputid2, panelid2, 'Text Input2', 'Panel2');
 
                 // remove instance and check label update
-                cy.get(`#${removeButton}`).find("button").click().then(() => {
-                    const [textinputid11, fieldView10] = Object.entries(formContainer._fields)[4];
-                    const [panelid11, fieldView11] = Object.entries(formContainer._fields)[5];
-                    checkLabelText(textinputid11, panelid11, 'Text Input2', 'Panel2');
+                // find Panel[1]'s remove button by its model index
+                const removeBtn1 = Object.values(formContainer._fields).find(f =>
+                    f.getModel &&
```
armaang1729, line 59: NOTE : 

Previous code used a hardcoded `_fields` array index to find the remove button, which broke after the instance removal logic was changed to id-based lookup (the field map order is no longer guaranteed to match insertion order). This change finds the remove button by inspecting the model — specifically the parent panel's `index === 1` — making the test resilient to field map reordering.

ui.tests/test-module/specs/repeatable/repeatable.updatelabel.cy.js
```diff
@@ -54,10 +54,27 @@ describe("Form Runtime with Panel Container", () => {
...
+                    f.getModel()?.fieldType === 'button' &&
+                    f.getModel()?.label?.value === 'Remove' &&
+                    f.getModel()?.parent?.index === 1
+                );
+                cy.get(`#${removeBtn1.getId()}`).find("button").click().then(() => {
+                    // after Panel[1] removed, find the surviving panel at index 1 by model
+                    const panel1remaining = Object.values(formContainer._fields).find(f =>
+                        f.getModel &&
```
armaang1729, line 67: NOTE : 

After removal, the surviving panel at model `index === 1` and its child text input are located dynamically rather than by a fixed `_fields` position. This is necessary because `#removeChildInstance` now splices by the live array index, which can shift the order of remaining entries in `formContainer._fields`.

armaang1729 (commented):
## 🤖 Code Review by Claude

Test review without inline comments

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/tableroweditorhook.js
```diff
@@ -0,0 +1,1108 @@
...
+    /**
+     * JSON for a new table header row (plain-text header cells), matching table _cq_template.
+     * @param {number} numCols
+     * @returns {object}
+     */
+    function buildHeaderRowContent(numCols) {
+        var baseTime = Date.now();
+        var content = {
+            "jcr:primaryType": "nt:unstructured",
+            "sling:resourceType": RESOURCE_TYPE_TABLEHEADER,
```
armaang1729, line 383: **🔴 Bug: `RESOURCE_TYPE_TABLEHEADER` is used but never declared — `ReferenceError` at runtime**

`buildHeaderRowContent()` references `RESOURCE_TYPE_TABLEHEADER` on this line, but only three resource-type constants are declared at the top of the file (`RESOURCE_TYPE_TABLEROW`, `RESOURCE_TYPE_TEXTINPUT`, `RESOURCE_TYPE_TEXT_DRAW`). Because the entire IIFE runs in `"use strict"`, accessing an undeclared binding throws a `ReferenceError` the first time an author clicks **Add Header Row**.

**Failure scenario:** Author selects a header row → clicks Add Header Row → `buildHeaderRowContent()` executes → strict-mode `ReferenceError` is thrown → `.fail()` handler never fires (the error surfaces as an uncaught exception) → no header row is created, no user-visible error message.

**Fix:**
```js
// add alongside the other RESOURCE_TYPE_* constants (lines 23-25)
var RESOURCE_TYPE_TABLEHEADER = "core/fd/components/form/tableheader/v1/tableheader";
```

ui.af.apps/src/main/content/jcr_root/apps/core/fd/af-clientlibs/core-forms-components-runtime-all/.content.xml
```diff
@@ -5,4 +5,4 @@
...
...core.forms.components.textinput.v1.runtime,core.forms.components.numberinput.v1.runtime,core.forms.components.panelcontainer.v1.runtime,core.forms.components.radiobutton.v2.runtime,core.forms.components.text.v1.runtime,core.forms.components.checkboxgroup.v1.runtime,core.forms.components.button.v1.runtime,core.forms.components.image.v1.runtime,core.forms.components.dropdown.v1.runtime,core.forms.components.fileinput.v4.runtime,core.forms.components.accordion.v1.runtime,core.forms.components.tabs.v1.runtime,core.forms.components.wizard.v1.runtime,core.forms.components.verticaltabs.v1.runtime,core.forms.components.recaptcha.v1.runtime,core.forms.components.checkbox.v1.runtime,core.forms.components.fragment.v1.runtime,core.forms.components.switch.v1.runtime,core.forms.components.termsandconditions.v1.runtime, core.forms.components.hcaptcha.v1.runtime,core.forms.components.review.v1.runtime,core.forms.components.turnstile.v1.runtime,core.forms.components.scribble.v1.runtime,core.forms.components.datetime.v1.runtime,core.forms.components.imagechoice.v1.runtime,core.forms.components.table.v1.runtime,core.forms.components.tablerow.v1.runtime,core.forms.components.tableheader.v1.runtime]"/>
```
armaang1729: **🔴 Regression: `checkboxgroup` runtime silently downgraded from v2 to v1**

While appending the three new table clientlib categories, `checkboxgroup.v2.runtime` was accidentally changed to `checkboxgroup.v1.runtime`. The v2→v1 upgrade was intentional (it fixed a screen-reader double-announcement bug where NVDA read both the `<fieldset>`/`<legend>` group context and a redundant per-input `aria-label`).

**Failure scenario:** Any form using a v2 checkboxgroup component loads `core.forms.components.runtime.all`; only the v1 JS runs. When a rule calls `setEnum`/`updateEnum` to add options dynamically, v1's `createCheckBoxItem` stamps `aria-label="GroupLabel: OptionLabel"` on each new `<input>`. Because the SSR structure is still `<fieldset><legend>`, NVDA announces the group name twice for every option, and any v2-only behaviour is silently absent.

**Fix:** restore the original v2 reference on this line:
```
core.forms.components.checkboxgroup.v2.runtime
```

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/tableroweditorhook.js
```diff
@@ -0,0 +1,1108 @@
...
+                return $.ajax({
+                    url: Granite.HTTP.externalize(itemPath),
+                    type: "POST",
+                    data: deleteParams
+                });
+            });
+        });
+
+        // Set the accumulated colspan on the surviving first cell
+        chain.then(function () {
```
armaang1729: **🟠 Bug: `chain.then(...)` result discarded — delete failures are silently swallowed in `mergeTableHeaderCells`**

The `forEach` loop (lines 964–972) builds up `chain` as a sequence of delete AJAX calls. On line 976, `chain.then(colspanPost).done(refresh).fail(notify)` creates a **new derived promise** but does **not** reassign it back to `chain`. This means:

- The `.fail(notify)` handler is attached to the derived promise, not to `chain`.
- If any delete in the forEach loop rejects, `chain` rejects — but there is no `.fail()` on `chain` itself, so the failure notification is never shown to the author.
- The colspan POST and the `tableEditable.refresh()` on line 983 are also silently skipped.

**Failure scenario:** Merging 3 header cells — the second delete AJAX call fails (network error or insufficient permissions) → `chain` rejects → the orphaned derived promise's `.fail()` never fires → author sees no error, the first cell is deleted, the colspan is never updated, and the table header is left in a partially-merged broken state.

**Fix:** assign the full chain back so the error handler is reachable:
```js
chain = chain.then(function () {
    return $.ajax({ url: ..., data: { colspan: String(totalColspan) } });
}).done(function () {
    if (tableEditable) { tableEditable.refresh(); }
}).fail(function () {
    author.ui.helpers.notify({ content: Granite.I18n.get("Failed to merge header cells."), ... });
});
```

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/js/tableview.js
```diff
@@ -0,0 +1,233 @@
...
+                if (!rowElement) {
+                    return;
+                }
+                const addBtn = rowElement.querySelector("[data-cmp-hook-add-instance]");
+                const removeBtn = rowElement.querySelector("[data-cmp-hook-remove-instance]");
+                if (addBtn) {
+                    addBtn.setAttribute(dataVisible, !(items.length === maxOccur && maxOccur !== -1));
+                }
+                if (removeBtn) {
+                    removeBtn.setAttribute(dataVisible, items.length > minOccur && minOccur !== -1);
```
armaang1729: **🟠 Bug: remove button permanently hidden when `minOccur` is not explicitly set in the model**

`minOccur` is an optional JCR property (`InjectionStrategy.OPTIONAL` in the Java model). When the author does not set it, the value is `undefined` in the JS model object.

The expression on this line evaluates as:
```js
items.length > undefined  // → false (NaN comparison)
  && undefined !== -1     // → true
// → false always
```

So the remove button's `data-cmp-visible` is set to `false` on every `#syncTableRowRepeatableControls` call, making it **impossible to remove rows** even when the table has far more rows than the implied minimum.

Note the asymmetry with the add-button check on line 221: `!(items.length === undefined && undefined !== -1)` → `!(false)` → `true`, so the add button is always visible, which is the opposite of the expected clamped behaviour.

**Fix:** treat missing/undefined as 0 for `minOccur` and -1 (unlimited) for `maxOccur`:
```js
const effectiveMin = (typeof minOccur === 'number' && minOccur >= 0) ? minOccur : 0;
const effectiveMax = (typeof maxOccur === 'number' && maxOccur >= 0) ? maxOccur : -1;
if (addBtn) {
    addBtn.setAttribute(dataVisible, !(items.length === effectiveMax && effectiveMax !== -1));
}
if (removeBtn) {
    removeBtn.setAttribute(dataVisible, items.length > effectiveMin);
}
```

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/js/tableview.js
```diff
@@ -0,0 +1,233 @@
...
+            } else {
+                const prevIndex = instanceIndex - 1;
+                const prevChild = children.find(c => c.getModel && c.getModel().index === prevIndex);
+                if (prevChild && prevChild.element) {
+                    prevChild.element.after(htmlElement);
+                } else {
+                    const items = instanceManager._model.items || [];
+                    const prevModel = items.find(m => m.index === prevIndex);
+                    if (prevModel) {
+                        const prevElement = tbody.querySelector(`#${prevModel.id}`);
```
armaang1729: **🟡 Bug: unescaped model ID in `querySelector` causes `SyntaxError` for IDs with CSS-special characters**

AEM Adaptive Forms model IDs are derived from JCR node names and qualified expressions. While typical node names are CSS-safe, IDs that include colons, brackets, or dots (e.g. `guide[0].table[0].row0[0]`) produce an **invalid CSS selector** when interpolated directly:

```js
// If prevModel.id = 'guide[0].table1[0].row0[0]'
tbody.querySelector('#guide[0].table1[0].row0[0]')
// → DOMException: SyntaxError — querySelector called with an invalid selector
```

The exception is unhandled; the insertion falls into neither `prevElement.after()` nor the `tbody.appendChild` fallback, so the new row is left **uninserted** and the table silently loses data.

**Fix:** use `CSS.escape()` (or the equivalent `getElementById` which accepts any id value):
```js
// Option A — CSS.escape (supported in all modern browsers):
const prevElement = tbody.querySelector(`#${CSS.escape(prevModel.id)}`);

// Option B — getElementById (no escaping needed, exact match):
const prevElement = document.getElementById(prevModel.id);
```

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/js/tableview.js
```diff
@@ -0,0 +1,233 @@
...
+
+            const instanceIndex = addedModel.index;
+            const children = instanceManager.children;
+
+            if (children.length === 0) {
+                tbody.appendChild(htmlElement);
+            } else if (instanceIndex === 0) {
+                const firstChild = children[0];
+                if (firstChild && firstChild.element) {
+                    tbody.insertBefore(htmlElement, firstChild.element);
```
armaang1729, line 149: **🟡 Bug: `insertBefore` with a stale/detached `firstChild.element` throws `HierarchyRequestError`**

When `instanceIndex === 0` and `children.length > 0`, this line calls:
```js
tbody.insertBefore(htmlElement, firstChild.element);
```

`instanceManager.children` holds view objects, not DOM references — there is no teardown path that nulls out `childView.element` after the view is removed from the IM. If a row was removed (view spliced from `children` in `#removeChildInstance`) but then re-added during the same model sync before the children array stabilises, `children[0]` may reference a view whose `.element` is already **detached from the DOM**.

Calling `insertBefore(newNode, detachedNode)` throws a `HierarchyRequestError` because `detachedNode` is no longer a child of `tbody`.

**Failure scenario:** A repeatable table starts with 1 row. The model emits a remove+add in the same tick (e.g., prefill replacing the initial row). The remove splices the view from `children`, but the DOM `<tr>` remains detached because `handleRemoval` ran first. The subsequent add calls `addRepeatableMarkup` with `children.length === 0` — this is actually fine in this specific case. However, if the timing differs and `children` still holds the stale view, `insertBefore` throws.

**Fix:** guard with `isConnected` before using a child's element as an anchor:
```js
const firstChild = children[0];
if (firstChild && firstChild.element && firstChild.element.isConnected) {
    tbody.insertBefore(htmlElement, firstChild.element);
} else {
    tbody.insertBefore(htmlElement, tbody.firstElementChild);
}
```

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableHeaderImpl.java
```diff
@@ -0,0 +1,42 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2024 Adobe
```
iamsudhanshu: Please fix across the PR - the copyright year.

iamsudhanshu:
Review comments from claude - @armaang1729 could you please review. 

---
  PR Review: #1879 — Added Table Component Support in Core Components
  
  Author: armaang1729 | State: Open (Work in Progress) | +6,059 / -33 across 78 files

  ---
  Overview

  This PR introduces a new Table component for AEM Adaptive Forms v2, including:
  - Sling Models: TableImpl, TableHeaderImpl, TableRowImpl
  - A large authoring editor hook (tableroweditorhook.js, 1437 lines)
  - Runtime clientlib registration
  - IT/example content with basic, columnwidth, mergesplit, and sorting variants
  - Column width calculation for both CSS and DOR (Document of Record) output
  - Authoring toolbar actions: add/delete row, add/delete column, move row, merge/split cells, sorting

  ---
  Critical Issues
  
  1. Duplicate constant with different names (ReservedProperties.java)
  public static final String PN_COLUMN_WIDTH = "columnWidth";
  public static final String PN_DOR_COLUMN_WIDTHS = "columnWidth";
  Both constants have identical values. This is misleading at best and a bug at worst. If they serve different purposes, they need different JCR property keys;
  if they're the same, one should be deleted and the other reused.
  
  2. RESOURCE_TYPE_TABLEHEADER used but never declared in tableroweditorhook.js
  In buildHeaderRowContent() (line ~1600 in the diff):
  "sling:resourceType": RESOURCE_TYPE_TABLEHEADER,
  The top of the file only declares RESOURCE_TYPE_TABLEROW, RESOURCE_TYPE_TEXTINPUT, and RESOURCE_TYPE_TEXT_DRAW. RESOURCE_TYPE_TABLEHEADER is never defined —
  this is a runtime ReferenceError in the "Add Header Row" flow.
  
  3. Unintentional clientlib downgrade
  In core-forms-components-runtime-all/.content.xml:
  - core.forms.components.checkboxgroup.v2.runtime
  + core.forms.components.checkboxgroup.v1.runtime
  This looks like an accidental regression — v2 was replaced with v1 for checkboxgroup. This will break checkbox group behavior for all forms using this
  aggregate clientlib.
  
  4. Javadoc misplaced on getExportedType() in TableImpl.java
  The Javadoc describing getColumnWidthColStyles() (explaining <col style="..."> and proportional width calculation) appears above getExportedType() at line
  183. The methods are in the wrong order relative to their documentation.

  ---
  Medium Issues

  5. Negative column widths silently pass validation
  In getColumnWidthColStyles():
  values[i] = Integer.parseInt(parts[i].trim()); // no check for negatives
  Integer.parseInt("-1") doesn't throw — it returns -1. A value like "-1,2,3" would pass through and generate a negative width: -33% CSS style. Add if 
  (values[i] < 0) values[i] = 0; or validate the input.
  
  6. Sequential AJAX chains are slow for wide tables
  Both addTableColumn and deleteTableColumn chain row updates sequentially with chain.then(...). For a table with 10+ rows each requiring a separate .1.json
  fetch + POST, this will be visibly slow. The row cell additions/deletions are independent — they should be parallelized with $.when(...promises).

  7. No dedicated Table interface
  TableImpl exposes isEnableSorting() and getColumnWidth() publicly via JSON but adapts only to Panel.class. There's no Table extends Panel interface for
  consumers to discover these properties through the public API surface. This is inconsistent with the pattern used by other components (e.g. Accordion, Tabs).

  8. Commented-out dead code in _cq_editConfig.xml
  The merge/split header cell actions are commented out in a block:
  
  Shipping commented-out features in a new component is a signal of incomplete implementation. These should either be removed entirely or tracked in a follow-up
  issue.

  9. filterComponent closure captures outer variables but signature was changed
  In replacehook.js, filterComponent and bindEventToReplaceComponentDialog were changed from accepting parameters to being parameter-less closures. However, if
  this hook is ever reused or the closure scope changes, this becomes fragile. The original parameter passing was safer.

  ---
  Minor Issues

  10. Copyright year inconsistency
  _cq_template.xml and tableroweditorhook.js use Copyright 2026 Adobe, while all other new files use Copyright 2024 Adobe. Pick one consistently.

  11. Mixed JUnit 4/5 assertions in TableImplTest
  import static org.junit.Assert.assertEquals;  // JUnit 4
  import static org.junit.Assert.assertNull;      // JUnit 4
  @ExtendWith(AemContextExtension.class)          // JUnit 5
  Use org.junit.jupiter.api.Assertions from JUnit 5 instead for consistency with the rest of the test suite.

  12. getColumnWidthColStyles() column widths don't sum to 100%
  Math.floor is used, so "1,1,1" → 33%+33%+33% = 99%. Consider rounding the last column to fill to 100%, or using Math.round strategically.

  13. Missing EOF newline in _cq_template.xml
  Flagged by the diff: \ No newline at end of file

  14. Trailing whitespace at end of replacehook.js
  +
  \ No newline at end of file
  Four spaces of trailing whitespace before the EOF.

  15. Test coverage gaps
  TableImplTest only has 2 tests for getDorProperties(). No tests for:
  - isEnableSorting() / getColumnWidthColStyles() / isColumnWidthsConfigured()
  - Invalid columnWidth values (e.g. "abc,1", "-1,2", "0,0")
  - TableHeaderImpl and TableRowImpl exported types

  ---
  PR Process Issues

  16. Empty PR description
  All sections (Description, Related Issue, Motivation, Testing, Checklist) are placeholder comments with no content filled in. No linked Jira/GitHub issue. The
  "work in progress" label confirms this isn't ready for review, but this should be filled before requesting review.

  ---

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
```diff
@@ -176,6 +176,19 @@ private ReservedProperties() {
 
     public static final String FD_DRAFT_ID = "fd:draftId";
 
+    public static final String PN_COLUMN_WIDTH = "columnWidth";
+
+    /** Comma-separated proportional column widths passed into fd:dor for DOR table rendering (matches GuideTableElement key). */
+    public static final String PN_DOR_COLUMN_WIDTHS = "columnWidth";
```
armaang1729, line 185: 🤖 **Reviewed by Claude**

**Duplicate constant — both map to the same string `"columnWidth"`**

`PN_COLUMN_WIDTH` and `PN_DOR_COLUMN_WIDTHS` are declared two lines apart and both resolve to the literal string `"columnWidth"`. This means:
1. `TableImpl.getDorProperties()` silently writes the same JCR key twice under different Java names — whichever write happens second wins, which is fine only because they are the same value, but the duplication makes the intent unclear and is a latent bug if either value is ever changed.
2. Any future refactor that renames one constant won't break the compiler, so the error will only surface at runtime.

**Fix:** Remove `PN_COLUMN_WIDTH` and use only `PN_DOR_COLUMN_WIDTHS` everywhere, or give the two constants genuinely different string values if they really map to different JCR properties.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableImpl.java
```diff
@@ -0,0 +1,134 @@
...
+    @Override
+    @NotNull
+    public String getExportedType() {
+        return ReservedProperties.VT_TABLE;
+    }
+
+    @Override
+    @JsonIgnore
+    @NotNull
+    public Map<String, Object> getDorProperties() {
```
armaang1729, line 94: 🤖 **Reviewed by Claude**

**Misplaced Javadoc — `getColumnWidthColStyles` comment decorates `getExportedType`**

The Javadoc block that says "One entry per column for `<col style="...">`" is placed immediately above `getExportedType()`, not above `getColumnWidthColStyles()` where it belongs. As written, every IDE and documentation generator will attribute that comment to `getExportedType`. Move the comment to sit directly above `getColumnWidthColStyles()`.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/tableroweditorhook.js
```diff
@@ -0,0 +1,1437 @@
...
+            return $.ajax({
```
armaang1729, line 980: 🤖 **Reviewed by Claude**

**`chain.then(...).done()` is detached from `chain` — colspan write failure is silently swallowed**

In both `mergeTableHeaderCells` and `mergeTableRowCells`, the pattern is:

```js
var chain = $.when();
// ... build delete chain ...
chain.then(function () {        // ← colspan POST
    return $.ajax(...);
}).done(function () {           // ← refresh
    tableEditable.refresh();
}).fail(function () {
    // error
});
```

The actual problem: if any intermediate delete AJAX call fails, jQuery `then`-chaining stops and the final colspan write is skipped, but the cell nodes deleted up to that point are already gone. The `.fail()` handler fires with a generic toast but gives the author no indication that the table is now in an inconsistent state (some cells deleted, no colspan written).

Consider surfacing a more specific warning when the fail handler fires after a partial delete sequence.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/tableroweditorhook.js
```diff
@@ -0,0 +1,1437 @@
...
+            yesBtn.on("click", function () {
+                var headerCellPath = headerPath + "/" + selectedCellName;
+
+                $.ajax({
+                    url: Granite.HTTP.externalize(headerCellPath),
+                    type: "POST",
+                    data: getDeleteParams()
+                }).then(function () {
+                    return $.ajax({
+                        url: Granite.HTTP.externalize(tablePath + ".1.json"),
```
armaang1729, line 804: 🤖 **Reviewed by Claude**

**Partial-failure: header cell is deleted before body cells, no rollback on fetch failure**

In `deleteTableColumn`, the chain is:
1. DELETE header cell
2. GET `tablePath.1.json` (to enumerate rows)
3. DELETE each body-row cell at `colIndex`

If step 2 fails (network error, 404, etc.), the table is left with the header column already deleted but all body cells still present. The single `.fail()` handler fires with a generic toast, giving the author no recovery path.

Consider fetching `tablePath.1.json` *before* any DELETE (before showing the dialog, or inside the dialog open), so by the time the user confirms, you already have the full list of paths to delete and can issue all DELETEs in one chain without the intermediate GET.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/table.html
```diff
@@ -0,0 +1,86 @@
...
+     id="${table.id}"
+     data-cmp-is="adaptiveFormTable"
+     data-cmp-adaptiveformcontainer-path="${formstructparser.formContainerPath}"
+     data-cmp-visible="${table.visible ? 'true' : 'false'}"
+     data-cmp-enabled="${table.enabled ? 'true' : 'false'}"
+     data-cmp-sorting-enabled="${properties.enableSorting ? 'true' : 'false'}">
+    
+    <sly data-sly-test="${table.label.visible != false}">
+        <div class="cmp-adaptiveform-table__title cmp-container__label">${table.label.value @ context='html'}</div>
+    </sly>
```
armaang1729, line 33: 🤖 **Reviewed by Claude**

**Label renders with `context='html'` unconditionally — should respect `label.richText`**

The shared `label.html` template uses `context = labelRichText ? 'html' : 'text'`. The table bypasses this template and always uses `context='html'`, meaning a non-rich-text label value is rendered as raw HTML rather than escaped text. While label content is author-controlled (not end-user input), keeping the context consistent with all other components avoids a category mismatch and future confusion.

Change to: `${table.label.value @ context=(table.label.richText ? 'html' : 'text')}`

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/editor/js/tableeditor.js
```diff
@@ -0,0 +1,30 @@
...
+(function(channel) {
+    "use strict";
+
+    /**
+     * Editor-specific initialization for table component
+     */
+    channel.on("cq-editor-loaded", function() {
+        console.log("Table component editor loaded");
+        
+        // Add any editor-specific logic here
```
armaang1729, line 26: 🤖 **Reviewed by Claude**

**Debug `console.log` left in production editor clientlib**

`tableeditor.js` ships a `console.log("Table component editor loaded")` and does nothing else. This file has no functional code — it is a stub. Either remove the file entirely (and drop it from `js.txt`) or populate it with real initialization logic. Shipping empty stubs with debug logs adds noise to every page load in the editor.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/tableroweditorhook.js
```diff
@@ -0,0 +1,1437 @@
...
+                        var rt = getResourceTypeFromTableJson(tableJson, rowName);
+                        if (rt !== RESOURCE_TYPE_TABLEROW) {
+                            return;
+                        }
+                        var rowPath = tablePath + "/" + rowName;
+                        (function (rp, rIndex) {
+                            chain = chain.then(function () {
+                                return fetchOrderedChildNames(rp).then(function (cellNames) {
+                                    if (colIndex < 0 || colIndex >= cellNames.length) {
+                                        return $.when();
```
armaang1729, line 715: 🤖 **Reviewed by Claude**

**`addTableColumn`: `colIndex` is the JCR child index, not the logical column index — breaks with merged header cells**

`colIndex` is computed as `headerOrder.indexOf(selectedCellName)` — the zero-based position among the header's *JCR child nodes*. When used to index into `cellNames` of each body row (which always stores one JCR node per logical cell), this is correct *only* when no header cell has a colspan > 1.

If a header cell with `colspan=2` exists at position 0, the selected cell at JCR position 1 represents logical column 3, but body rows still have one JCR child per *logical* column. The resulting `colIndex=1` would insert the new body cell after position 1, which is out of sync with where the header column was added.

Consider computing `colIndex` as the *logical* column offset (sum of colspans up to the selected header cell) rather than the raw JCR child index.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/js/tableview.js
```diff
@@ -0,0 +1,348 @@
...
+         * @param {HTMLTableSectionElement} tbody
+         * @param {HTMLTableSectionElement} thead
+         * @param {number} colIndex
+         */
+        #sortTableByColumn(tbody, thead, colIndex) {
+            const rows = Array.from(tbody.querySelectorAll(":scope > tr"));
+            if (rows.length <= 1) {
+                return;
+            }
+            let dir = "asc";
```
armaang1729, line 271: 🤖 **Reviewed by Claude**

**Sort reorders `InstanceManager.children` but not the AF model — breaks subsequent add/remove**

`#syncInstanceManagerOrderAfterSort` reorders `im.children` to match the new DOM order. However, the AF model's `items` array (and each row's `model.index`) remains in the original, unsorted order.

When `addRepeatableMarkup` subsequently runs (e.g. the user adds a row after sorting), it uses `addedModel.index` and looks up `children.find(c => c.getModel().index === prevIndex)` — but `children` is now in sorted (visual) order, while `model.index` still reflects the pre-sort position. This can cause the new row to be inserted at the wrong position, or `children.find()` to return the wrong view.

Since the AF runtime owns the model order, a client-only visual sort that mutates `im.children` without a corresponding model reorder is structurally unsafe. Consider either: (a) not mutating `im.children` at all and using pure CSS/DOM sort that the IM doesn't track, or (b) issuing a model reorder event so the AF model updates to match.

ui.tests/test-module/libs/commons/formsConstants.js
```diff
@@ -52,7 +52,10 @@ var formsConstants = {
                 "termsandconditions": "/apps/forms-components-examples/components/form/termsandconditions",
                 "submitButton": "/apps/forms-components-examples/components/form/actions/submit",
                 "review": "/apps/forms-components-examples/components/form/review",
-                "scribble": "/apps/forms-components-examples/components/form/scribble"
+                "scribble": "/apps/forms-components-examples/components/form/scribble",
+                "table": "/apps/forms-components-examples/components/form/table",
+                "tablerow": "/apps/forms-components-examples/components/form/tablerow",
```
armaang1729, line 57: 🤖 **Reviewed by Claude**

**`tablerow` and `tableheader` paths in `formsConstants.js` point to non-existent example components**

`"tablerow": "/apps/forms-components-examples/components/form/tablerow"` and the `tableheader` entry reference example-layer component paths, but this PR only adds an example component for `table` (at `examples/ui.apps/.../form/table/`). No example components for `tablerow` or `tableheader` are included.

Any IT test that uses these constants to locate a component (e.g. to insert it via the component browser) will silently fail or throw a 404. Either add the missing example component overlays or remove these entries from `formsConstants.js` and use the core resource paths directly in the test helpers.

armaang1729:
🤖 **Reviewed by Claude — Code Review Summary**

### Critical Issues

**1. Duplicate constant in `ReservedProperties.java`** (inline comment posted)
`PN_COLUMN_WIDTH` and `PN_DOR_COLUMN_WIDTHS` both equal `"columnWidth"`. One should be removed or they need different string values.

**2. Misplaced Javadoc on `getExportedType()` in `TableImpl.java`** (inline comment posted)
The comment describing `getColumnWidthColStyles()` is attributed to `getExportedType()`. IDEs and Javadoc generators will misattribute it.

**3. Sort desync: `#syncInstanceManagerOrderAfterSort` mutates `IM.children` without updating the AF model** (inline comment posted)
After a client-side sort, `InstanceManager.children` is reordered but `model.items` and each row's `model.index` remain in the original order. A subsequent add-row or remove-row will insert at the wrong position.

**4. `addTableColumn` uses JCR child index, not logical column index** (inline comment posted)
When header cells have `colspan > 1`, `colIndex` diverges from the logical column position in body rows. Adds a body cell at the wrong slot.

**5. `deleteTableColumn` partial-failure leaves the table inconsistent** (inline comment posted)
The header cell is deleted first, then body-row cells are deleted one by one. A network failure after the first DELETE leaves the table in a broken state with no recovery path.

**6. `mergeTableHeaderCells`/`mergeTableRowCells`: partial delete + no colspan write on chain failure** (inline comment posted)
If any intermediate cell DELETE fails, already-deleted cells are gone but colspan is never written. The generic `.fail()` toast gives no indication of the inconsistent table state.

### Structural Issues

**7. `tablerow` and `tableheader` entries in `formsConstants.js` point to non-existent example components** (inline comment posted)
IT tests using these constants will fail.

**8. `tableeditor.js` is an empty stub with a `console.log`** (inline comment posted)
No functional code; remove or populate.

**9. Label uses `context='html'` unconditionally in `table.html`** (inline comment posted)
The shared label template conditionalizes on `richText`. The table always uses `context='html'`, departing from the established pattern.

### Coverage Gap

`TableHeaderImpl` and `TableRowImpl` have no unit tests. At minimum, `getExportedType()` should be covered to catch accidental reversion to the resource-type default.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/js/tableroweditorhook.js
armaang1729: removed the merge header cells functionality

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableHeaderImpl.java
armaang1729: done

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/js/tableview.js
armaang1729: Resolved

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/base/v1/base/_cq_editConfig.xml
```diff
@@ -29,6 +29,62 @@
...
+                hidden="{Boolean}true"
+                icon="tableMergeCells"
+                text="Merge Cells"/>
+        <splitheadercell
+                jcr:primaryType="nt:unstructured"
+                condition="function(editable){ return CQ.FormsCoreComponents.editorhooks.isMergedHeaderCell(editable); }"
+                handler="CQ.FormsCoreComponents.editorhooks.splitTableHeaderCell"
+                hidden="{Boolean}true"
+                icon="tableCellsSplit"
+                text="Split Cell"/> -->
```
sakshi-arora1, line 59: can we get rid of the commented part.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/css/tableview.css
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+/* All visual styles live in the theme canvas:
+ * aem-forms-theme-canvas/src/components/table/_table.scss
+ */
+
+.cmp-adaptiveform-table {}
```
sakshi-arora1, line 21: Also raise a PR in canvas theme for adding table styles

sakshi-arora1 (approved):

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/clientlibs/site/css/tableview.css
armaang1729: Done, released a new canvas theme version [1.0.57](https://www.npmjs.com/package/@aemforms/af-canvas-theme/v/1.0.57) PR: https://github.com/adobe/aem-forms-theme-canvas/pull/106.

PR #1883, opened 2026-05-15
feat(af-core): FORMS-25463 expose cq:annotations parallel to fd:dor

prateek1497:
Adds cq:annotations read support to the IC GET flow for the print channel /
   author CRISPR path.

  - getCqAnnotations() reads the cq:annotations child resource (a sibling of fd:dorContainer on the component node) and returns each annotation as a map entry keyed by node name. Each entry carries color, text, x, y, and optional state/resolvedBy/resolvedAt plus JCR audit fields. Null fields are stripped so the serialized shape matches the JCR content exactly.
  - getProperties() is wired to call getCqAnnotations() and place the result under properties.cq:annotations — parallel to fd:dor / fd:path / fd:associate. Returns null when no cq:annotations child exists; Jackson omits the key, keeping output non-breaking for forms without annotations.
  - getCustomProperties() explicitly excludes cq:annotations to prevent it leaking as a double-serialized JSON string via the catch-all custom-properties path.
  - Gated: only runs when channel == PRINT and isAuthorMode(request) — web channel and publish mode are unaffected.

## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
@@ -314,10 +314,6 @@ public class AbstractFormComponentImpl extends AbstractComponentImpl implements 
 
     public static final String CUSTOM_RULE_PROPERTY_WRAPPER = "fd:rules";
 
-    /**
-     * Key under which reviewer annotations (cq:annotations) are exposed in the component's serialized properties map,
-     * parallel to fd:dor / fd:path.
-     */
     public static final String CUSTOM_ANNOTATIONS_PROPERTY_WRAPPER = "cq:annotations";
 
     /**
@@ -752,33 +748,13 @@ public class AbstractFormComponentImpl extends AbstractComponentImpl implements 
     }
 
     /**
-     * Returns the reviewer annotations for this component, read from the {@code cq:annotations} child resource on the
-     * component node (a sibling of {@code fd:dorContainer}). Each annotation node's properties are serialized into a
-     * per-annotation map keyed by node name; the result is placed parallel to {@code fd:dor} under the component's
-     * properties.
-     *
-     * <p>
-     * Scoped to authoring/review IC print forms only; runtime/publish requests must not depend on this authoring
-     * payload. Same print-channel gate as {@link #getDorContainer()}, plus an explicit publish-view denylist via
-     * {@link FormConstants#REQ_ATTR_PUBLISH_VIEW} so the runtime publish path never receives this data.
-     *
-     * @return ordered map keyed by annotation node name with the annotation's properties (text, x, y, state,
-     *         reviewedBy, reviewedAt, resolvedBy, resolvedAt, ...); or null when the channel is not print, the
-     *         resource is missing, the request is in publish view, or no cq:annotations child exists. Returning
-     *         null lets callers omit the property and keeps the output non-breaking for forms without annotations.
+     * Returns the reviewer annotations stored under the component's {@code cq:annotations} child resource,
+     * keyed by annotation node name. Scoped to the print channel; returns null when no annotations exist
+     * so callers can omit the property.
      */
     @JsonIgnore
     public Map<String, Object> getCqAnnotations() {
-        // Gate matches getDorContainer() (print channel + resource present) plus an explicit
-        // publish-view denylist. The previous isAuthorMode(request) check relied on WCMMode,
-        // but the IC REST GET (CommunicationReadProcessor → FormModelReader) resolves the Sling
-        // Model with a wrapped request that doesn't carry WCMMode, so the gate always failed
-        // and cq:annotations never reached the response. See the pre-existing TODO above about
-        // sling-model-wrapper requests having incorrect WCMMode for the same reason.
         if (!FormConstants.CHANNEL_PRINT.equals(this.channel) || resource == null) {
-            return null;
-        }
-        if (request != null && Boolean.TRUE.equals(request.getAttribute(FormConstants.REQ_ATTR_PUBLISH_VIEW))) {
             return null;
         }
         Resource annotationsResource = resource.getChild("cq:annotations");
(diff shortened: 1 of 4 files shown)


prateek1497:
In short why its not following AFv2/sites annotations - because it does network call for each annotation, bringing the infinity json for that annotation, for IC, we would add cq:annotations in the first GET call, that is adding it in authoring crispr, and flushing everything on the UI at once. and same in the existing SAVE call of Delrina, all annotation would get save at once. We have made changes at CoreComponentFormGenerator for this

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -740,4 +747,31 @@ public Map<String, Object> getDorContainer() {
...
+        }
+        Resource annotationsResource = resource.getChild("cq:annotations");
+        if (annotationsResource == null) {
+            return null;
+        }
+        Map<String, Object> result = new LinkedHashMap<>();
+        for (Resource child : annotationsResource.getChildren()) {
+            ValueMap vm = child.getValueMap();
+            Map<String, Object> ann = vm.entrySet().stream()
+                .filter(e -> isAllowedType(e.getValue()))
```
rismehta: **Warning — JCR internal properties leak into annotation output**

`isAllowedType` accepts `String`, so `jcr:primaryType = "nt:unstructured"` passes the filter. Every annotation node will have this key in its exported map. Other JCR audit fields (`jcr:created`, `jcr:createdBy`, `jcr:modified`, etc.) stored as `Calendar`/`String` will also leak.

Mirror the prefix exclusion from `getCustomProperties()`:
```java
.filter(e -> isAllowedType(e.getValue())
    && !e.getKey().startsWith("jcr:")
    && !e.getKey().startsWith("sling:"))
```
Also add a test assertion: `assertNull(annotation.get("jcr:primaryType"))`.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
```diff
@@ -643,6 +649,7 @@ private Map<String, Object> getCustomProperties() {
             .stream()
             .filter(entry -> isAllowedType(entry.getValue())
                 && !reservedProperties.contains(entry.getKey())
+                && !CUSTOM_ANNOTATIONS_PROPERTY_WRAPPER.equals(entry.getKey())
```
rismehta: **Warning — `ReservedProperties` convention bypassed**

Per `docs/architecture/overview.md` §8.2, every JCR property name must be registered in `ReservedProperties.java` — `aggregateReservedProperties()` auto-collects all `String` fields. The PR adds a one-off inline guard instead.

Add to `ReservedProperties.java`:
```java
public static final String PN_CQ_ANNOTATIONS = "cq:annotations";
```
Then the inline filter here becomes unnecessary (once the constant is in `ReservedProperties`, `reservedProperties.contains("cq:annotations")` is true and the existing guard covers it).

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImplTest.java
```diff
@@ -273,6 +274,57 @@ public void testPrintChannelRule() {
...
+        Map<String, Object> properties = abstractFormComponentImpl.getProperties();
+        Map<String, Object> annotations = (Map<String, Object>) properties.get("cq:annotations");
+        assertNotNull(annotations, "cq:annotations should appear for print channel");
+        Map<String, Object> annotation = (Map<String, Object>) annotations.get("annotation-1");
+        assertNotNull(annotation);
+        assertEquals("review note", annotation.get("text"));
+    }
+
+    @Test
+    public void testAnnotationPropertiesIncludeAllAllowedTypes() {
```
rismehta: **Warning — test name misrepresents coverage**

`testAnnotationPropertiesIncludeAllAllowedTypes` only exercises `String` and `Long`. `isAllowedType` also accepts `Boolean`, `Calendar`, `BigDecimal` and their array variants — none are covered. Rename to `testAnnotationPropertiesIncludeStringAndLongTypes`, or add fixture data and assertions for at least one additional type.

rismehta (commented):
## Code Review — PR #1883: expose cq:annotations parallel to fd:dor

### Summary

Solid feature addition with good channel-scoping logic and defensive null checks. Two **Warnings** need fixing before merge, plus a few suggestions.

---

### Positive
- Good early-return guards (null resource, non-print channel).
- Correct Mockito stub added to `testAssociateProperties` — necessary because STRICT_STUBS mode causes failure if `resource.getChild("cq:annotations")` is invoked without a stub.
- `LinkedHashMap` preserves annotation node order.

---

### Suggestions
- **`getCqAnnotations()` visibility**: method is `public` with no interface declaration. Analogous `getAssociateProperties()` is `private`. Change to `private` unless a subclass override use case is documented.
- **Missing edge-case test**: no test covers the case where a `cq:annotations` child node exists but has zero annotation children (method returns null — branch untested).
- **`reviewedBy`/`reviewedAt` in fixture**: these fields appear in the test fixture but are not listed in the PR description schema. Either assert them or clarify.
- `if (annotations != null && !annotations.isEmpty())` in `getProperties()` is redundant since `getCqAnnotations()` already returns null for empty — minor cleanup.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
prateek1497: added this

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImpl.java
prateek1497: added this

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/util/AbstractFormComponentImplTest.java
prateek1497: fixed this

rismehta (approved):

PR #1888, opened 2026-05-23
Table component second check in 

armaang1729:
## Types of changes

## Checklist:


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableHeaderImpl.java
```diff
@@ -0,0 +1,42 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2024 Adobe
```
devgurjar, line 2: Change it to 2026

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableImpl.java
```diff
@@ -0,0 +1,134 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2024 Adobe
```
devgurjar, line 2: Change it to 2026

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableRowImpl.java
```diff
@@ -0,0 +1,42 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2024 Adobe
```
devgurjar, line 2: Change it to 2026

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableImplTest.java
```diff
@@ -0,0 +1,93 @@
+/*~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ ~ Copyright 2024 Adobe
```
devgurjar, line 2: Change it to 2026

devgurjar:
Please add tests cases to fix the code coverage

devgurjar (commented):
Reviewed against the existing AEM Forms (cq-guides) table component at `/libs/fd/af/components/table/` for feature parity and against the AEM Core Components patterns for code style. Overall the architecture is solid (clean separation of TableImpl/HeaderImpl/RowImpl, sensible edit-vs-publish HTL split with the foster-parent comment, DoR property pass-through). Most of the comments below are correctness or polish.

### Feature parity vs the existing GuideTable in cq-guides

| Existing GuideTable | This PR | Notes |
|---|---|---|
| `columnWidth` | (/) `columnWidth` | same JCR property |
| `sortingEnabled` | (/) **`enableSorting`** | property renamed -- existing forms migrating to v1 won't carry over without a content upgrade hook |
| `colspan` per header cell | (/) via `cell.valueMap.colspan` | |
| `disableSorting` per header cell | (/) via `cell.valueMap.disableSorting` | |
| `mobileLayout` (responsive) | (x) | existing has `tableMobileLayout.less` + `responsiveTablePlugin.js` -- worth flagging as roadmap item |
| `tableSummaryText` (a11y) | (x) | modern equivalent is `<caption>` (inline comment on table.html) |
| `minOccur`/`maxOccur` per row | (x) | existing used these for repeatable bounds; only `repeatable` boolean is in the new model |
| `breakBefore/AfterText`, `overflowText` | (x) | likely intentional -- XFA-print-specific, DoR templates handle this |
| `css` (custom class) | (x) | |

### Scope / process notes (not inline because the diff doesn't have a clean anchor)

- **No dialog UI for the new properties.** `_cq_dialog/.content.xml` shows `+0/-3` -- 3 lines removed, no new field added for `columnWidth` or `enableSorting`. Authors can only set these by editing JCR directly. If the dialog is a deliberate follow-up check-in, please call that out in the PR body.
- **No README for the parent `table` component covering the new properties.** READMEs for `tablerow` and `tableheader` were added. The `table` README was modified `+8/-8` -- worth confirming it documents the new `columnWidth`/`enableSorting`.
- **PR description is the empty template.** Worth filling in: a link to the JIRA, what the "first check-in" PR is (so cumulative diff is reviewable), and the explicit out-of-scope items (mobile layout, summary, dialog).
- (For completeness: the `text/_cq_editConfig.xml` change is **not** unrelated as it first appears -- it adds tableheader-cell context actions (merge/split/enable/disable sort) gated by `condition` callbacks. Could still go in a separate PR for cleaner review, but it's intentional.)

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableImpl.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    @JsonIgnore
+    public boolean isColumnWidthsConfigured() {
+        return !getColumnWidthColStyles().isEmpty();
+    }
+
+    /**
+     * One entry per column for {@code <col style="...">}: {@code width: N%} (no HTL string concatenation).
+     * For example, proportional {@code "1,1,4"} becomes {@code width: 16%}, {@code width: 16%}, {@code width: 66%}.
+     */
```
devgurjar, line 88: **Detached Javadoc -- belongs to a different method.**

The Javadoc block on lines 82-85 describes `<col style="width: N%">` rendering and the proportional-width math ("1,1,4 -> 16%, 16%, 66%"). But it's anchored above `getExportedType()`, which has nothing to do with column widths -- it just returns the constant `VT_TABLE`.

The Javadoc looks like it was originally above `getColumnWidthColStyles()` (line 109) and got displaced during an edit. Move it back to its real method so the next reader doesn't have to chase it.

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableImpl.java
```diff
@@ -0,0 +1,134 @@
...
+                values[i] = 1;
+            }
+            sum += values[i];
+        }
+        if (sum == 0) {
+            return Collections.emptyList();
+        }
+        List<String> result = new ArrayList<>();
+        for (int v : values) {
+            int pct = (int) Math.floor((v * 100.0) / sum);
```
devgurjar, line 129: **Two issues in `getColumnWidthColStyles()`:**

**1. Floor on the percent can lose up to (cols-1)% total**

- `"1,1,4"` -> 16% + 16% + 66% = **98%**
- `"1,1,1"` -> 33% + 33% + 33% = **99%**

Combined with the inline `table-layout: fixed; width: 100%` in `table.html:64`, columns will visually misalign by a percent or two. The standard fix is to round the first N-1 columns and let the last absorb the rounding remainder:

```java
int remaining = 100;
for (int i = 0; i < values.length - 1; i++) {
    int pct = (int) Math.round((values[i] * 100.0) / sum);
    result.add("width: " + pct + "%");
    remaining -= pct;
}
result.add("width: " + remaining + "%");
```

**2. Silent fallback to `1` on `NumberFormatException` (line 119-120)**

If an author types `"1,foo,2"`, they get a misaligned table with no diagnostic anywhere -- the bad token silently becomes a 1-unit column. At minimum log a warning so the author can find the typo; alternatively skip the bad token entirely and return an empty list (forcing the even-split fallback).

Worth adding a test for: `getColumnWidthColStyles()` with malformed input, all-zero values, and the rounding-remainder case (`"1,1,1"` should sum to 100%).

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
```diff
@@ -176,6 +176,19 @@ private ReservedProperties() {
 
     public static final String FD_DRAFT_ID = "fd:draftId";
 
+    public static final String PN_COLUMN_WIDTH = "columnWidth";
+
+    /** Comma-separated proportional column widths passed into fd:dor for DOR table rendering (matches GuideTableElement key). */
+    public static final String PN_DOR_COLUMN_WIDTHS = "columnWidth";
```
devgurjar, line 182: **Two constants with identical string values.**

```java
public static final String PN_COLUMN_WIDTH       = "columnWidth";   // line 179
public static final String PN_DOR_COLUMN_WIDTHS  = "columnWidth";   // line 182
```

Same literal `"columnWidth"` under two different names. Either:

- These are semantically different (e.g. the DoR-exported property might want to be `"dorColumnWidths"` or similar) -- in which case the values should differ; or
- They really are the same key -- in which case there should be one constant.

As-is, this is a footgun: a future refactor that changes one value won't touch the other, and the two callers (`TableImpl.columnWidth` field annotation and `getDorProperties()`) will silently diverge. If the intent was to use the same JCR property both as the local read source and as the DoR pass-through key, drop `PN_DOR_COLUMN_WIDTHS` and use `PN_COLUMN_WIDTH` in both places with a comment explaining.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/table.html
```diff
@@ -25,7 +25,8 @@
      data-cmp-is="adaptiveFormTable"
      data-cmp-adaptiveformcontainer-path="${formstructparser.formContainerPath}"
      data-cmp-visible="${table.visible ? 'true' : 'false'}"
-     data-cmp-enabled="${table.enabled ? 'true' : 'false'}">
+     data-cmp-enabled="${table.enabled ? 'true' : 'false'}"
+     data-cmp-sorting-enabled="${properties.enableSorting ? 'true' : 'false'}">
```
devgurjar, line 29: **Template reads `properties.*` directly, bypassing the Sling Model.**

Line 29:
```html
data-cmp-sorting-enabled="${properties.enableSorting ? 'true' : 'false'}"
```

Also line 64 (in the publish-mode block):
```html
data-sly-attribute.style="${properties.columnWidth ? 'table-layout: fixed; width: 100%' : false}"
```

The Sling Model already exposes these as `${table.enableSorting}` and `${table.columnWidthsConfigured}`. Reading raw `properties.*` sidesteps the model -- e.g. the `Boolean.TRUE.equals(enableSorting)` null-check in `TableImpl.isEnableSorting()` doesn't apply here, and any future change to how that property is computed (delegation, mixin, fallback to a parent) won't take effect through the raw read.

Use the model accessors consistently:
```html
data-cmp-sorting-enabled="${table.enableSorting ? 'true' : 'false'}"
data-sly-attribute.style="${table.columnWidthsConfigured ? 'table-layout: fixed; width: 100%' : false}"
```

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/table.html
```diff
@@ -25,7 +25,8 @@
...
     
     <sly data-sly-test="${table.label.visible != false}">
         <div class="cmp-adaptiveform-table__title cmp-container__label">${table.label.value @ context='html'}</div>
```
devgurjar, line 32: **Accessibility -- prefer `<caption>` over a `<div>` title.**

Currently the title is rendered as a sibling div before the `<table>`:
```html
<div class="cmp-adaptiveform-table__title cmp-container__label">${table.label.value @ context='html'}</div>
...
<table aria-label="${table.label.value @ context='attribute'}">
```

For screen readers, the modern recommendation is to put the title **inside** the table as a `<caption>` element:
```html
<table>
    <caption class="cmp-adaptiveform-table__title">${table.label.value @ context='html'}</caption>
    ...
</table>
```

`<caption>` is automatically read by assistive tech as the table's title and is structurally bound to it -- the duplicated `aria-label` becomes unnecessary, and you avoid having the title visually associated with the table but semantically detached. Existing GuideTable used `<table summary>` which is deprecated in HTML5; `<caption>` is the modern replacement.

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/table/v1/table/table.html
```diff
@@ -36,21 +37,50 @@
...
-            <sly data-sly-list.child="${resource.listChildren}">
-                <sly data-sly-test="${child.resourceType == 'core/fd/components/form/tableheader/v1/tableheader'}">
-                    <sly data-sly-resource="${child @ decoration=false}"></sly>
+    <!--/* EDIT MODE: column widths not applied; children render their own <tr> with data-cq-data-path for the overlay. */-->
+    <sly data-sly-test="${wcmmode.edit}">
+        <table class="cmp-adaptiveform-table__widget"
+               aria-label="${table.label.value @ context='attribute'}">
+            <thead class="cmp-adaptiveform-table__head">
+                <sly data-sly-list.child="${resource.listChildren}">
+                    <sly data-sly-test="${child.resourceType == 'core/fd/components/form/tableheader/v1/tableheader'}">
```
devgurjar, line 46: **Hardcoded resource-type strings duplicated in HTL.**

Lines 46, 53, 72, 79 all do:
```html
<sly data-sly-test="${child.resourceType == 'core/fd/components/form/tableheader/v1/tableheader'}">
<sly data-sly-test="${child.resourceType == 'core/fd/components/form/tablerow/v1/tablerow'}">
```

The constants `RT_FD_FORM_TABLE_HEADER_V1` and `RT_FD_FORM_TABLE_ROW_V1` exist in `FormConstants.java` but the HTL hardcodes the strings -- if those constants ever change (e.g. v2 of header/row), this template won't follow.

The common AEM Core Components pattern is to expose **per-child predicates** from the Sling Model on the parent:

```java
// In TableImpl
public boolean isTableHeader(Resource child) {
    return child != null && FormConstants.RT_FD_FORM_TABLE_HEADER_V1.equals(child.getResourceType());
}
```

Then the HTL becomes `${table.tableHeader(child)}` and the resource-type string lives in one place. (Or define an enum-style children list on the model so the template doesn't need predicates at all.)

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/tableheader/v1/tableheader/tableheader.html
```diff
@@ -14,7 +14,8 @@
   ~ limitations under the License.
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~*/-->
 <sly data-sly-use.header="com.adobe.cq.forms.core.components.models.form.Panel"
-     data-sly-use.formstructparser="com.adobe.cq.forms.core.components.models.form.FormStructureParser"></sly>
+     data-sly-use.formstructparser="com.adobe.cq.forms.core.components.models.form.FormStructureParser"
+     data-sly-set.sortingEnabled="${resource.parent.valueMap.enableSorting}"></sly>
```
devgurjar, line 18: **Implicit-coupling to the parent's resource shape.**

```html
<sly data-sly-set.sortingEnabled="${resource.parent.valueMap.enableSorting}"></sly>
```

This assumes the table-header's direct parent is always the table itself. Fine today, but the moment the structure changes -- a wrapper panel for layout grouping, a CSP migration that re-nests components, an MSM/inherited-config indirection -- this breaks silently and produces a header that ignores sort configuration.

Cleaner alternatives:

1. **Expose it on `TableHeaderImpl`**: walk up to find the nearest `RT_FD_FORM_TABLE_V1` ancestor in Java, expose `${header.sortingEnabled}`. The walk handles wrapper layers explicitly.
2. **Use the `FormStructureParser`** (already loaded on line 17): if it knows the form structure, add a `getEnclosingTableSortingEnabled(headerResource)` method.

Not blocking for v1, but worth a TODO comment at minimum.

bundles/af-core/src/test/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TableImplTest.java
```diff
@@ -0,0 +1,93 @@
...
+            dorProps.get("columnWidth"));
+    }
+
+    @Test
+    void testGetDorProperties_columnWidthPassedThroughWhenAuthored() throws Exception {
+        Panel table = Utils.getComponentUnderTest(PATH_TABLE_WITH_COLUMN_WIDTH, Panel.class, context);
+        Map<String, Object> dorProps = table.getDorProperties();
+        assertEquals("columnWidth must carry the authored proportional value",
+            "1,2,1", dorProps.get("columnWidth"));
+    }
```
devgurjar, line 92: **Test coverage gap -- two tests for a 135-line model.**

The two tests cover only `getDorProperties()` column-width pass-through. The following methods have zero coverage and several have real branches worth pinning:

- `isEnableSorting()` -- null / `Boolean.TRUE` / `Boolean.FALSE` (the `Boolean.TRUE.equals(...)` null-guard is invisible otherwise)
- `getColumnWidth()` -- null vs configured
- `isColumnWidthsConfigured()` -- both branches
- `getColumnWidthColStyles()` -- happy path, malformed input (`"1,foo,2"`), all-zero (`"0,0,0"`), single column, rounding remainder (`"1,1,1"` should sum to 100%, see comment on `TableImpl.java:129`)
- `getExportedType()` -- pin the constant return

Also missing: `TableHeaderImplTest`, `TableRowImplTest`. Those classes are tiny (just override `getExportedType()`) but a basic adapt-and-export check guards against the resource-type wiring breaking silently.

The test fixture (`test-content.json`) has only two table nodes with no header/row children -- so nothing exercises the parent-child wiring that's central to this component. Add a richer fixture with `core/fd/components/form/tableheader/v1/tableheader` + `core/fd/components/form/tablerow/v1/tablerow` child nodes.

PR #1896, opened 2026-05-29
FORMS-25687: Adobe Sign Block component (master)

Ikshu-Jain-1:


ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/.content.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <jcr:root xmlns:cq="http://www.day.com/jcr/cq/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0"
     jcr:primaryType="cq:ClientLibraryFolder"
-    categories="[core.forms.components.container.editor.hook.v2,cq.authoring.editor.hook]"/>
+    categories="[core.forms.components.container.editor.hook.v2,cq.authoring.editor.hook]"
+          embed="[aem.adobesign.rteplugin,core.forms.components.adobesign.placeholder]"/>
```
devgurjar: 🟡 **[important] This embeds the Adobe Sign RTE plugin into the *shared* container editor-hook clientlib, loaded for every adaptive form being authored — not just forms containing an Adobe Sign Block.**

I compared this against how the other optional/addon components (`recaptcha`, `hcaptcha`, `turnstile`) are wired: none of them embed into this shared clientlib. Each keeps its own clientlib self-contained and loads it on demand via `extraClientlibs` on its own `_cq_dialog`, only when that component's dialog is actually opened.

This new component already does exactly that correctly — `adobesignblock/v1/adobesignblock/clientlibs/editor/.content.xml` declares `dependencies="[aem.adobesign.rteplugin, core.forms.components.base.v1.editor, core.forms.components.adobesign.placeholder]"`, and its `_cq_dialog` pulls it in via `extraClientlibs="[core.forms.components.adobesignblock.v1.editor]"`. So this `embed=` addition on the shared container clientlib looks redundant with that on-demand path, and its effect is that every adaptive form's authoring session now loads the Adobe Sign RTE plugin JS, even for forms that never use this component.

(The `core.forms.components.adobesign.placeholder` half is probably harmless — that clientlib is also registered under `cq.authoring.editor.forms`/`cq.authoring.editor.forms.v2`, which appear to load globally on every forms-editor page already, so AEM's clientlib de-dup likely makes that part a no-op. The RTE plugin is the one that adds genuinely new, unconditional payload.)

Is there a reason `toolbaractionhook.js`'s `isAdobeSignBlock` check needs the RTE plugin loaded globally, ahead of that component's own dialog being opened? If not, dropping this `embed=` and relying on the existing `extraClientlibs` path would keep the bundle-loading pattern consistent with the other optional components.

devgurjar (commented):
One architecture/performance finding from review — everything else looked solid (well-documented dialog/model code, DOMPurify sanitization consistent with existing textview.js/FormFieldBase.js convention, correct HTL rich-text rendering).

ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/container/v2/container/clientlibs/editorhook/.content.xml
Ikshu-Jain-1: Addressed

devgurjar (approved):

PR #1898, opened 2026-06-08
chore(skill): close create-core-component gaps surfaced by Date Input

im-shiv:
## Description

Closes gaps in the `create-core-component` skill that were surfaced while building and reviewing the Date Input component (PR #1881). Every defect found in that component traced back to a missing or actively-misleading instruction in the skill — this PR updates the skill (workflow, references, validator) so future components don't repeat them.

## Gaps identified & addressed

**1. Runtime view-JS override contract was missing / wrong (root cause of most runtime bugs)**
- The accessibility checklist instructed overriding `updateValidity`/`updateEnabled`/`updateReadOnly` to *only* set ARIA/`disabled`, with no mention of calling `super` — which silently drops the base behaviour (`data-cmp-*` attributes, error-message rendering, `--filled`/`--empty`).
- Added `references/runtime-view-js.md` documenting the full `FormFieldBase` override contract: call `super.<handler>(value, state)` first; an overridden `updateValue` must call `updateEmptyStatus()`; `setModelValue` is not echoed back to the originating field.
- Rewrote the checklist's "JS View Methods" table and SKILL.md Phase 4 accordingly.

**2. No guidance for composite / split widgets**
- The focus-clobber bug (typing `2024` → `024`) and stray submitted sub-field params came from having no pattern for multi-input widgets.
- Added a "Composite / split widget" base-class category and a dedicated section: one hidden combined input carries the value + `name`; never put `name` on visible sub-inputs; never repopulate a sub-input the user is actively editing.

**3. Cypress tests were entirely absent from the skill**
- The skill only covered unit tests + exporter JSON, so HTL/view-JS/dialog/sync bugs were never caught.
- Added `references/cypress-tests.md` (required runtime + authoring specs, IT content, `formsConstants.js` entry, runtime-all embed) plus the two Cypress gotchas: hidden inputs fail `be.visible`; chained `should('(not.)have.attr')` breaks because the first yields the attribute value.

**4. Broken `aria-labelledby` pattern**
- The shared label template emits no `id`, so `aria-labelledby="...__label"` references a non-existent element. Documented; grouped widgets must use `aria-label`.

**5. BEM modifier classes with no CSS**
- `hideTitleDate` shipped a `--hidden` class that had no styling (dead feature). Added the rule: every emitted modifier must have a matching CSS rule; never ship an empty design dialog.

**6. Editor-JS safety & i18n**
- Use `textContent` not `innerHTML`; don't wrap locale-neutral format strings in `Granite.I18n.get`; wrap hard-coded default UI text with `@ i18n`.

**7. Incomplete/incorrect file checklist (`component-anatomy.md`)**
- Was missing the Cypress specs, `_cq_styleConfig`, the runtime-all embed and the `formsConstants.js` entry; had the wrong IT-content path and listed `.less` instead of `.css`. Corrected.

**8. Validator (`scripts/validate_component.py`) extended with 4 new check categories**
- view-JS override contract (`super` on `update*`, `updateEmptyStatus` in `updateValue`)
- editor-JS `innerHTML` usage
- BEM modifier classes have a CSS rule + empty-design-dialog warning
- Cypress specs / `formsConstants` entry / runtime-all embed existence + Cypress-gotcha lints

## Types of changes

- [x] Skill / tooling improvement (no runtime component code changed)

### DIFF
--- a/.claude/skills/create-core-component/SKILL.md
+++ b/.claude/skills/create-core-component/SKILL.md
@@ -78,41 +78,54 @@ Read `references/templates.md` for the full template set. Read `references/compo
    - Label container with label + question mark
    - Widget area (component-specific)
    - Short description, long description, error message sections
+   - Wrap any **hard-coded default UI text** (e.g. fallback sub-labels) with `@ i18n`
+   - **Composite widgets**: put `name` only on the **hidden combined input**, never on the visible sub-inputs (avoids stray submitted params); find sub-inputs in JS by class
+   - Any **BEM modifier class** you emit (e.g. `--hidden`) MUST have a matching CSS rule — a class with no style is a no-op feature
    - **CRITICAL**: Read `references/accessibility-checklist.md` and ensure every applicable item is satisfied
 
 6. **Renderer JS** — Return paths to shared field templates
 
 7. **Dialog XML** — Include base dialog fields via `granite/ui/components/coral/foundation/include`, add component-specific tabs/fields
 
-8. **Site clientlib** — `.content.xml`, `js.txt`, `css.txt`, view JS, LESS styles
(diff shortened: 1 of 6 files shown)


armaang1729 (commented):
LGTM, Thanks for updating these skills.

PR #1906, opened 2026-06-17
feat(form-container): own fd:changeEventBehaviour via node property + feature toggle

rismehta:
## Summary

Brings `fd:changeEventBehaviour` fully into the core component. Two related changes:

### 1. Honor `fd:changeEventBehaviour` set on the node
`AbstractFormComponentImpl.getCustomProperties()` filters out every `fd:`/`jcr:`/`sling:` prefixed node property, so a value set directly on the form container node never reached `model.json`. Registered it in `ReservedProperties` and added a dedicated `@ValueMapValue` getter on `FormContainerImpl`.

### 2. Drive the toggle-default from the core component (replaces cq-guides addon)
`fd:changeEventBehaviour="deps"` was injected by the cq-guides `CoreComponentCustomPropertiesProvider`, which read the Granite toggle `FT_FORMS-12053` directly. The core component now supports feature toggles, so this moves in-tree: `FormContainerImpl` sets it when `FT_ALLOW_MULTIPLE_FIELDS_IN_WHEN` (`FT_FORMS-12053`) is enabled, read via `ComponentUtils.isToggleEnabled` (system property). The `@OSGiService CoreComponentCustomPropertiesProvider` reference is removed.

**Precedence:** toggle provides the default; a value on the node overrides it.

## Coordinated PRs (sequencing matters)
- **aemds**: bridge Granite toggle `FT_FORMS-12053` to a system property via `ToggleMonitorSystemPropertyFactory`, and bump af-core to this release (must ship with/before the cq-guides removal).
- **cq-guides**: remove the now-unused `CoreComponentCustomPropertiesProvider` interface, impl and test.

No-regression note: during transition both the old provider and this code may emit the same `"deps"` value (idempotent). The only unsafe order is removing the cq-guides provider before the aemds system-property bridge exists.

## Tests
- `testGetChangeEventBehaviourFromNode`, `testNodeChangeEventBehaviourOverridesToggleDefault`
- `testChangeEventBehaviourFromToggle`, `testChangeEventBehaviourAbsentWhenToggleDisabled`
- `testJSONExportWithChangeEventBehaviour` — full JSON exporter test (mirrors `testJSONExportWithAutoSaveEnable`) asserting `fd:changeEventBehaviour="deps"` is present in the exported model JSON, with new `test-content-change-event-behaviour.json` + `exporter-formcontainerv2ChangeEventBehaviour.json` fixtures. Also passes schema validation.

64 v2 `FormContainerImplTest` cases pass.

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
@@ -40,6 +40,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.adobe.aemds.guide.common.GuideContainer;
+import com.adobe.aemds.guide.service.CoreComponentCustomPropertiesProvider;
 import com.adobe.aemds.guide.service.GuideSchemaType;
 import com.adobe.aemds.guide.utils.GuideConstants;
 import com.adobe.aemds.guide.utils.GuideUtils;
@@ -95,6 +96,9 @@ public class FormContainerImpl extends AbstractContainerImpl implements FormCont
 
     /** Constant representing spreadsheet submit action type */
     private static final String SS_SPREADSHEET = "spreadsheet";
+
+    @OSGiService(injectionStrategy = InjectionStrategy.OPTIONAL)
+    private CoreComponentCustomPropertiesProvider coreComponentCustomPropertiesProvider;
 
     @OSGiService(injectionStrategy = InjectionStrategy.OPTIONAL)
     private HttpClientBuilderFactory clientBuilderFactory;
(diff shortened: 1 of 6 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v2/form/FormContainerImpl.java
```diff
@@ -400,6 +403,11 @@ public String getLanguageDirection() {
             }
         }
         properties.put(FD_ROLE_ATTRIBUTE, getRoleAttribute());
+        // A value set on the node overrides the toggle-driven default above (consistent with how other
+        // node-level custom properties take precedence).
+        if (StringUtils.isNotBlank(changeEventBehaviour)) {
```
sakshi-arora1, line 419: This will not override the behaviour if at system level the toggle was disabled, this conflicts with the comment

rismehta: Good catch — reworded the comment. A value on the node now documents that it **always** takes precedence and is emitted regardless of the system-level toggle/provider state (when the toggle is disabled there's no default above, so it simply sets the node-authored value). Behaviour itself is unchanged — node value always wins. Also restored the `CoreComponentCustomPropertiesProvider.getProperties()` block, so the addon and the in-tree toggle both supply the `deps` default idempotently.

PR #1923, opened 2026-07-17
feat: add asyncValidtionExpression property to form field/container model

vdua:
## Summary
- Adds `asyncValidtionExpression` (JCR property name intentionally matches, typo included) to the form field/container Sling Model, mirroring the existing `validationExpression` property.
- Wired at the single shared point (`AbstractBaseImpl`), so it's automatically exposed on every `Field`/`Container` subtype (text input, number input, date picker, panel, etc.) without per-component overrides.
- Allow-listed in the `0.15.2` JSON schema fixtures used by tests, in the same 11 field-type blocks that already permit `validationExpression`.
- `RuleUtils.js` now also scans `asyncValidtionExpression` for custom function references (same treatment as `validationExpression`), so custom functions referenced only in async validation expressions still get registered on `window`.

Out of scope (explicitly, per discussion): a paired constraint-message property (`validateExpMessage`-style) and any authoring/dialog UI — this PR only adds the raw expression property to the model and schema.

## Changes
- **Java model**: `BaseConstraint.java` (default method), `ReservedProperties.java` (JCR property constant), `AbstractBaseImpl.java` (`@ValueMapValue` field + `@JsonInclude(NON_EMPTY)` getter)
- **Schema**: `adaptive-form-data-constraints.schema.json`, `adaptive-form.schema.json` (0.15.2)
- **Java tests**: `BaseConstraintTest`, `TextInputImplTest` (blank-value suppression + populated round-trip export/schema validation)
- **JS**: `RuleUtils.js` + `RuleUtils.test.js`
- **Docs**: note in `docs/architecture/overview.md` on deriving golden-file `id`s from a failing-test diff instead of hand-computing them

## Test plan
- [x] `TextInputImplTest` — 62/62 passing (`mvn surefire:test -Dtest=TextInputImplTest`)
- [x] `BaseConstraintTest` — 4/4 passing
- [x] `RuleUtils.test.js` — 12/12 passing (`npx jest RuleUtils.test.js`)

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
@@ -59,6 +59,7 @@ public final class ReservedProperties {
     public static final String PN_DOR_TEMPLATE_REF = "dorTemplateRef";
     public static final String PN_DOR_TYPE = "dorType";
     public static final String PN_VALIDATION_EXPRESSION = "validationExpression";
+    public static final String PN_ASYNC_VALIDATION_EXPRESSION = "asyncValidtionExpression";
     public static final String PN_REQUIRED = "required";
     public static final String PN_AUTOCOMPLETE = "autocomplete";
     public static final String PN_ASSIST_PRIORITY = "assistPriority";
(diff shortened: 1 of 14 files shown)


bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/form/ReservedProperties.java
```diff
@@ -59,6 +59,7 @@ private ReservedProperties() {
     public static final String PN_DOR_TEMPLATE_REF = "dorTemplateRef";
     public static final String PN_DOR_TYPE = "dorType";
     public static final String PN_VALIDATION_EXPRESSION = "validationExpression";
+    public static final String PN_ASYNC_VALIDATION_EXPRESSION = "asyncValidtionExpression";
```
rismehta, line 62: Typo

rismehta:
@vdua You need to update the spec as well

PR #1927, opened 2026-07-22
FORMS-26630: Add show as popup property in guidecontainer model json

akshayvas:
## Types of changes

## Checklist:

### DIFF
--- a/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TermsAndConditionsImpl.java
+++ b/bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TermsAndConditionsImpl.java
@@ -46,6 +46,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 public class TermsAndConditionsImpl extends PanelImpl implements TermsAndConditions {
 
     private static final String CUSTOM_TNC_PROPERTY = "fd:tnc";
+    private static final String SHOW_AS_POPUP_PROPERTY = "fd:showAsPopup";
 
     @JsonIgnore
     @ValueMapValue(injectionStrategy = InjectionStrategy.OPTIONAL, name = ReservedProperties.PN_SHOW_APPROVAL_OPTION)
@@ -87,6 +88,7 @@ public class TermsAndConditionsImpl extends PanelImpl implements TermsAndConditi
         Map<String, Object> properties = super.getProperties();
         if (resource.getValueMap().containsKey(CUSTOM_TNC_PROPERTY)) {
             properties.put(CUSTOM_TNC_PROPERTY, true);
+            properties.put(SHOW_AS_POPUP_PROPERTY, isShowAsPopup());
         }
         return properties;
     }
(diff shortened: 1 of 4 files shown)


sakshi-arora1 (commented):
@akshayvas Where is the implementation honouring this configuration?

bundles/af-core/src/main/java/com/adobe/cq/forms/core/components/internal/models/v1/form/TermsAndConditionsImpl.java
```diff
@@ -87,6 +88,7 @@ public boolean isShowAsPopup() {
         Map<String, Object> properties = super.getProperties();
         if (resource.getValueMap().containsKey(CUSTOM_TNC_PROPERTY)) {
             properties.put(CUSTOM_TNC_PROPERTY, true);
+            properties.put(SHOW_AS_POPUP_PROPERTY, isShowAsPopup());
```
sakshi-arora1: This ties `fd:showAsPopup` export to the presence of the `fd:tnc` resource property, which represents an unrelated concept (custom T&C content marker). `isShowAsPopup()` is independent of whether `fd:tnc` is set.

In practice `fd:tnc` is stamped by the component's `_cq_template` on newly authored nodes, but any existing/legacy T&C content authored before that (or content created without going through the template) won't have `fd:tnc` on the resource. For that content, `fd:showAsPopup` will simply be missing from the exported JSON even when "Show as popup" is checked in the dialog, so the runtime has no way to know it should render as a popup.

Suggest moving this line (and probably the whole `showAsPopup` property) outside the `if (resource.getValueMap().containsKey(CUSTOM_TNC_PROPERTY))` block so it's always emitted, e.g.:
```java
properties.put(SHOW_AS_POPUP_PROPERTY, isShowAsPopup());
if (resource.getValueMap().containsKey(CUSTOM_TNC_PROPERTY)) {
    properties.put(CUSTOM_TNC_PROPERTY, true);
}
```

sakshi-arora1 (commented):
Overview: adds `fd:showAsPopup` to the exported properties of `TermsAndConditionsImpl` so consumers of the JSON model can tell whether the T&C panel should render as a popup. Note the PR title mentions "guidecontainer" but the diff only touches `TermsAndConditionsImpl` — worth fixing the title/description for traceability.

Main concern is inline below: the new property is gated behind an unrelated existing condition (`fd:tnc` presence), which will silently drop `fd:showAsPopup` from the export for any content that doesn't have that marker property set.

Test coverage: both updated fixtures (`exporter-termsandconditions.json`, `exporter-termsandconditionsNoWrapData.json`) and the underlying `test-content.json` already have `fd:tnc: true` hardcoded, so there's no test exercising the branch where `fd:tnc` is absent — i.e. no test proves/documents whether `fd:showAsPopup` is expected to disappear in that case.

akshayvas:
> @akshayvas Where is the implementation honouring this configuration?

@sakshi-arora1 
Implementation In headless - https://github.com/adobe/aem-forms-headless-components/pull/179

akshayvas:
@sakshi-arora1 I have addressed review comments in below PR, closing this, as current PR is failing to sync branch from fork. It will have only exporter json baseline test failure due to addition of new property.

https://github.com/adobe/aem-core-forms-components/pull/1929

PR #1938, opened 2026-07-30
FORMS-25942: Adding skills for foundation to core component migration and adding any custom component

Ikshu-Jain-1:
### What
Refactors and expands two Claude skills:
- `create-core-component` — scaffolding a new custom AEM Core Form Component
- `migrate-foundation-component` — foundation → core component migration

Bulk content is moved out of SKILL.md into on-demand reference files. SKILL.md is reduced to terse pointers and Critical Rules.

### Notes
- Reference files load only when needed, so they add no baseline context cost.
- All references are linked directly from SKILL.md (one level deep).

### Testing
I had used this skill while building the Adobe Sign components.


armaang1729:
## What the PR gets right

  The structural direction is sound — moving bulk content out of SKILL.md into reference files is exactly the progressive disclosure pattern the docs prescribe. common-mistakes.md is the strongest addition: project-specific failure modes Claude cannot infer from general knowledge (wrong fieldType format, data-sly-test on __value, @Inject vs @ValueMapValue) are high signal, low noise. The new HTL templates (7b, 7c, 2b, 6b) fill real gaps for display/text and step components that had no prior template coverage. The aria-describedby fix in template #7 corrects a real accessibility bug — old template had aria-label on the <input> widget which is the wrong element. The wiring steps as a numbered workflow checklist follows the docs' recommended pattern exactly for complex multi-step operations where skipping a step silently breaks things. Refactoring Critical Rules to be terse pointers to reference files rather than inline explanations is the right call — keeps SKILL.md scannable.

## Core Problem
  PR adds 1766 lines / 7 new reference files with zero justification. Empty PR body, no linked issues, no observed failure evidence. Inverts the best-practices mandate: "build evaluations first, then write minimal instructions."

  ---
  ### Issue 1 — No evidence-based rationale (Blocking)
  Docs require: identify observed failures → baseline → minimal fix → measure. PR has none of this. Cannot evaluate whether any of the 1766 lines actually improves skill performance or just adds noise.

  ---
 ### Issue 2 — Verbosity kills determinism (Blocking)
  More context = more paths for Claude to go off-script = hallucinations. PR documents the full component lifecycle instead of the creation task scope. Over-constraining an open-field task with exhaustive reference material makes Claude second-guess itself. Docs: "Only add context Claude doesn't already have. Challenge each piece."

  ---
  ### Issue 3 — Out-of-scope content (css-architecture.md)
  CSS stub creation = empty BEM rules, nothing else. css-architecture.md (70 lines) brings in aem-forms-theme-canvas, Parcel builds, design tokens, npm run create-clientlib — a post-finalization workflow owned by a separate team/process. Loads unnecessary context on every component creation run.
  
  ---
###  Issue 4 — Second-level reference nesting
  composite-multifield.md and common-mistakes.md both link to editor-clientlib.md. Chain: SKILL.md → X.md → editor-clientlib.md. Docs explicitly warn Claude uses head -100 on second-level refs and gets incomplete content.
  
   ---

  documentation referred : https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices

armaang1729 (changes requested):
Changes requested !

Ikshu-Jain-1:
Thanks for sharing the feedback.

**Issue 1:**
I’ve added a PR description for the same. 

**Issue 2:** 
The best-practices doc frames this as "set appropriate degrees of freedom," matched to task fragility. Since component scaffolding is a complex task, exhaustive reference material is required. Also, these files would only loaded on demand and would not consume context tokens if not required. I've also tried to ensure that we do not have any duplications with respect to context and we refer to the already existing context as much as possible.

**Issue 3:**
Addressed

**Issue 4:**
In this case editor-clientlib.md, common-mistakes.md and composite-multifield.md are a first-level reference linked directly from SKILL.md. The links between them are lateral "also see" pointers to the canonical source to avoid duplicating the JS boilerplate. It is not a nesting chain.

sakshi-arora1 (commented):
(comment retracted)

sakshi-arora1 (approved):

PR #1941, opened 2026-07-31
Show custom acceptMessage for unsupported file type in File Attachment

mayankjindalmj:
Original PR #1928 couldn't be merged due to a fork. 

## Summary
- The file-input widget's `invalidMessage()` looked up `constraintMessages.invalidMimeType` for the "unsupported file type" alert, but `invalidMimeType` is not an authorable constraint message key — the schema only exposes `constraintMessages.accept` (backed by the `acceptMessage` dialog property, see `AbstractBaseImpl.ConstraintMessages#getAcceptConstraintMessage`).
- As a result, the custom message authors configure for file-type validation was always ignored in favor of the generic fallback string, while the analogous `maxFileSize` custom message worked correctly since its key matched (`constraintMessages.maxFileSize`).
- Fixed in both the webpack source (`ui.frontend/src/view/FormFileInputWidgetBase.js`) and the apps clientlib copy (`ui.af.apps/.../fileinput/v4/fileinput/clientlibs/site/js/fileinputwidget.js`).

## Root cause (reproduced live)
| File uploaded | Alert before fix | Alert after fix |
|---|---|---|
| Oversized file (allowed type) | Custom `maxFileSize` message ✅ | unchanged ✅ |
| Unsupported file type | Generic fallback message ❌ (ignored author's `acceptMessage`) | Custom `accept`/`acceptMessage` text ✅ |

## Test plan
- [x] `node --check` on both modified JS files
- [x] Added Cypress test `should display a custom error message configured by the user for unsupported file type` in `fileinputv4.runtime.cy.js`, mirroring the existing `maxFileSize` custom-message test
- [x] Added `acceptMessage` to the `fileinput7` fixture field in the IT sample content used by that spec
- [ ] Full Cypress IT run against a live AEM instance (not run in this environment — no AEM instance available)

### DIFF
--- a/it/content/src/main/content/jcr_root/content/forms/af/core-components-it/samples/fileinput/fileinputv4/basic/.content.xml
+++ b/it/content/src/main/content/jcr_root/content/forms/af/core-components-it/samples/fileinput/fileinputv4/basic/.content.xml
@@ -108,6 +108,7 @@
                     jcr:title="File Input - 7"
                     sling:resourceType="core/fd/components/form/fileinput/v4/fileinput"
                     accept="[audio/*, video/*, image/*, text/*, application/pdf,.ifc]"
+                    acceptMessage="This file type is not supported!"
                     fieldType="file-input"
                     readOnly="{Boolean}false"
                     name="fileinput7"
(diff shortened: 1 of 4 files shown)


mayankjindalmj:
Closing in favor of the dev-targeted PR (same fix + test, cherry-picked).

PR #1970, opened 2026-08-26
Create core component skill

sakshi-arora1:
## Types of changes

## Checklist:

### DIFF
--- a/.claude/skills/migrate-foundation-component/references/property-mapping.md
+++ b/.claude/skills/migrate-foundation-component/references/property-mapping.md
@@ -1,0 +1,217 @@
+# Foundation → Core Property Mapping
+
+## Class Hierarchy Mapping
+
+| Foundation Class | Core Base Class | Core Interface | Component Type |
+|-----------------|-----------------|----------------|----------------|
+| `GuideField` | `AbstractFieldImpl` | `Field` | Simple field |
+| `GuideField` (with enum/options) | `AbstractOptionsFieldImpl` | `OptionsConstraint` | Options field |
+| `GuideTextBox` | `AbstractFieldImpl` | `Field` + `StringConstraint` | Simple field |
+| `GuideCheckBox` | `AbstractFieldImpl` | `Field` (boolean) | Simple field |
+| `GuideRadioButton` | `AbstractOptionsFieldImpl` | `OptionsConstraint` | Options field |
+| `GuideDropDownList` | `AbstractOptionsFieldImpl` | `OptionsConstraint` | Options field |
+| `GuideDatePicker` | `AbstractFieldImpl` | `Field` + `DateConstraint` | Simple field |
+| `GuideNumericBox` | `AbstractFieldImpl` | `Field` + `NumberConstraint` | Simple field |
+| `GuideFileUpload` | `AbstractFieldImpl` | `Field` + `FileConstraint` | Simple field |
+| `GuideButton` | `AbstractBaseImpl` | `Base` | Display-only (action) |
+| `GuideTextDraw` | `AbstractBaseImpl` | `Base` | Display-only |
+| `GuideSeparator` | `AbstractBaseImpl` | `Base` | Display-only |
+| `GuideImage` | `AbstractBaseImpl` | `Base` | Display-only |
+| `GuidePanel` | `AbstractContainerImpl` | `Container` | Container |
+| `GuideContainer` | `AbstractContainerImpl` | `Container` | Container |
+| `GuideCaptcha` | `AbstractFieldImpl` | `Field` | Simple field |
+
+## FieldType Mapping
+
+| Foundation `guideFieldType` | Core `FieldType` enum | FieldType string |
+|----------------------------|-----------------------|------------------|
+| `GUIDE_FIELD_TEXTBOX` | `TEXT_INPUT` | `text-input` |
+| `GUIDE_FIELD_TEXTBOX` (multiLine) | `MULTILINE_INPUT` | `multiline-input` |
+| `GUIDE_FIELD_NUMERIC_BOX` | `NUMBER_INPUT` | `number-input` |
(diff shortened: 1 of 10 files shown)


rismehta (changes requested):
Is this intended for customers or internal form developers? Have we validated it with the intended users?

Can you summarize the validation gate? Specifically, how are we validating the AI-generated output, and what mechanisms are in place for the AI to detect mistakes and self-correct?

Have we written evals for this skill?

.claude/skills/migrate-foundation-component/references/property-mapping.md
```diff
@@ -0,0 +1,247 @@
...
+| `wrapData` | `wrapData` | XFA data wrapping; not applicable | Nothing | No action needed |
+| `allowedParents` | (metadata) | Foundation parsys constraint; Core uses policy | **Template policy** — component groups in `wcm/policies` | Add component group to template policy |
+
+## Properties Requiring Per-Migration Review
+
+These properties exist in some foundation components and need case-by-case evaluation:
+
+| Foundation Property | Core Status | What to do |
+|--------------------|-------------|-----------|
+| `assistPriority` enum values | Supported, but enum values differ: Foundation uses `caption`, `toolTip`, `name`, `custom`; Core uses `LABEL`, `DESCRIPTION`, `NAME`, `CUSTOM` | Remap: `caption`→`LABEL`, `toolTip`→`DESCRIPTION`, `name`→`NAME`, `custom`→`CUSTOM` |
```
rismehta: We should not maintain a copy here, it should be re-used. This would be difficult to maintain

.claude/skills/migrate-foundation-component/references/property-mapping.md
```diff
@@ -0,0 +1,247 @@
...
+| `GUIDE_FIELD_FORM` | `FORM` | `form` |
+| (no equivalent) | `CHECKBOX_GROUP` | `checkbox-group` |
+
+## Common Property Mapping — GuideField → Field/Base
+
+These properties exist in the foundation base `GuideField` / `GuideNode` and are already handled by Core base classes (`AbstractFieldImpl`, `AbstractBaseImpl`). They do NOT need custom migration:
+
+| Foundation Property | JCR Name | Core Equivalent | Handled By |
+|--------------------|----------|-----------------|------------|
+| `getTitle()` / `jcr:title` | `jcr:title` | `getLabel().getValue()` | `AbstractBaseImpl` |
```
rismehta: It should refer to component documentation, we should maintain another list here

sakshi-arora1: Done. Removed the duplicated Core getter/JCR column. The table no longer restates the Core-side property surface — it points to the authoritative sources (the base-interface/implementation hierarchy in `create-core-component/references/component-anatomy.md`, and `Base.java`/`Field.java`) and keeps only the foundation-name → owning-base-class mapping, which isn't documented elsewhere.

.claude/skills/migrate-foundation-component/references/property-mapping.md
sakshi-arora1: Done. Removed the per-component "Core Equivalent" tables that duplicated each component's dialog properties. The section now directs the reader to the authoritative `README.md` for each Core component (`ui.af.apps/.../form/{component}/v{n}/{component}/README.md`) and retains only the foundation-specific transformation knowledge a README can't provide (renames, `items`→`enum`/`enumNames` restructuring, dropped/computed props). `SKILL.md` Phase 3 now instructs the AI to cross-check against the live README instead of a hardcoded copy.

sakshi-arora1:
Thanks for the review — responses to the three questions below.

### Intended for customers or internal form developers? Validated with intended users?

These skills are intended for **external customers customizing/authoring their own Adaptive Form components**. The goal is to give customers a guided, convention-correct path to scaffold a new Core Form Component (or migrate a Foundation/Guide component they own) without reverse-engineering the repo's Java/HTL/dialog/clientlib patterns by hand — the skill encodes those patterns, WCAG 2.1 AA accessibility requirements, and BEM naming so customer-authored components stay consistent with the Core Components framework.

Because the audience is external, a couple of implications we're being explicit about:
- **No assumed internal access** — the flow must work against a customer's own project checkout and a standard public Maven build, not internal-only tooling.
- **Versioning/support expectations** — customers rely on the base interfaces and each component's `README.md` as the stable contract, which is exactly why the property mappings now reference those sources instead of embedding a copy that could drift out from under customer components.

Validation with the intended (external) users hasn't happened yet — so far it's been internal dogfooding on existing components. I'll track structured validation with customer/partner developers as a follow-up before we position this as customer-ready.

### Validation gate — how is AI-generated output validated, and how does the AI self-correct?

Output goes through a layered gate rather than being trusted as-is:

1. **Static validator** — `create-core-component/scripts/validate_component.py` runs 90+ checks across 12 categories (file completeness, naming/BEM conventions, dialog structure, accessibility/WCAG patterns), reporting pass/fail per check.
2. **Build + unit verification** — `mvn clean compile` followed by the component's `…ImplTest` unit tests.
3. **Migration parity audits (Phase 7)** — feature-parity, dialog-parity, rendering-parity, and accessibility-upgrade audits comparing the migrated Core component against the foundation source.

**Self-correction loop:** the validator and build/test steps emit concrete, itemized failures; the workflow feeds those back so the AI fixes the flagged files and re-runs the gate until it's clean rather than proceeding on a first pass. The migration plan is also presented to the developer for confirmation before code generation (migrate Phase 5), keeping a human checkpoint in the loop.

### Have we written evals?

Not yet — there are no formal eval fixtures today. Current QA is the validation gate above (static validator + build/tests + parity audits) plus the human plan-approval checkpoint. A repeatable eval harness (fixture foundation component → run skill → assert output passes `validate_component.py` + build) is a sensible follow-up; capturing it as a TODO rather than expanding this PR's scope.

PR #1977, opened 2026-09-02
feat(recaptcha): support reCAPTCHA v3 in adaptive forms (+ XFA no-iframe embed fixes)

kartikey19427:
## Summary

Adds reCAPTCHA **v3** support to the adaptive forms recaptcha component, plus two fixes that make auto-fetch captchas work on XFA-backed forms embedded without an iframe.

## reCAPTCHA v3 (feature)

- Client-side widget rendering and edit-dialog logic treat v3 the same as an invisible/score-based key (v3 has no visible challenge).
- `customFunctions.js` fetches the v3 token locally (classic, non-enterprise `grecaptcha` namespace), since upstream `fetchCaptchaToken` only handles turnstile and Enterprise score-based keys.
- `submitForm()` is overridden to auto-fetch the captcha token before submit for v3 as well, matching the existing turnstile / Enterprise behavior.
- For XFA-rendered forms, the token is applied via `dispatchEvent(field, 'custom:setProperty', {value})` instead of a direct `field.value =` assignment, to work with af-core-xfa's get-only rule-node proxy.

## XFA no-iframe embed fixes

While validating v3 on an embedded form, two pre-existing bugs surfaced that broke **auto-fetch captcha** (invisible v2, enterprise-score, turnstile, and v3) on XFA-backed forms embedded **without an iframe**:

1. **Register custom functions into the XFA runtime at bundle load** (`ui.frontend/src/index.js`) — the runtime ships two bundles (non-XFA `@aemforms/af-core` and XFA `@aemforms/af-core-xfa`), each with its own `FunctionRuntime`. Custom functions were registered only into the non-XFA runtime via `setupFormContainer` (through the single `window.FormView`), so the XFA form's rule engine never received `fetchCaptchaToken` and its built-in `submitForm` failed with `fetchCaptchaToken is not defined`. Registering at bundle load populates each bundle's own runtime before the form model is built. `registerFunctions` is idempotent, so the existing registration is unaffected.
2. **Guard container-view submit/save handlers against a missing `state.events`** (`formcontainerview.js`) — the XFA runtime's `getState()` omits `events`, which crashed the submit/save success/error handlers (`Cannot read properties of undefined`) once submit progressed. Guarded with optional chaining; a missing events object means no rule is configured, so the default handler runs. No-op for the non-XFA runtime.

## Testing

Verified end-to-end on a local AEM instance with an enterprise-score reCAPTCHA on an XFA form embedded no-iframe: the XFA rule engine exposes `fetchCaptchaToken`, the token is fetched and set on submit, `POST /adobe/forms/af/submit/... -> 200 OK`, and the thank-you / success handling runs. No `fetchCaptchaToken is not defined` and no `submitSuccess` handler crash. Frontend unit tests pass.


ui.frontend/src/customFunctions.js
```diff
@@ -20,6 +20,146 @@ import * as cf from "@aemforms/af-custom-functions";
...
+ * apps for (it has no default `fetchCaptchaToken` at all), so this is not a hack - custom
+ * function-table entries are looked up dynamically by name and always take precedence over
+ * defaults.
+ *
+ * @param {...*} args - Author-supplied arguments (payload/validateForm/contentType, or the
+ *  deprecated success/error/contentType/data/validateForm form), followed by the globals object
+ *  the framework always appends last.
+ * @returns {object} - Empty object, matching the original's return value.
+ */
+function submitForm() {
```
rismehta: submitForm should not be overridden, we should use a pre-submit hook if required ? Also, can you check old captcha implementation

ui.frontend/src/customFunctions.js
```diff
@@ -20,6 +20,146 @@ import * as cf from "@aemforms/af-custom-functions";
...
+    if (a === null || a === undefined) {
+        return a;
+    }
+    if (isArrayValue(a)) {
+        return a.map(function (i) { return valueOf(i); });
+    }
+    return a.valueOf();
+}
+
+function toStringOrEmpty(a) {
```
rismehta: These functions are not required, make them inline or private

ui.frontend/src/customFunctions.js
```diff
@@ -20,6 +20,146 @@ import * as cf from "@aemforms/af-custom-functions";
...
+ * Delegates to @aemforms/af-custom-functions for turnstile and reCAPTCHA Enterprise (unchanged
+ * behavior). reCAPTCHA v3 is handled locally: the upstream package's fetchCaptchaToken only knows
+ * about the grecaptcha.enterprise namespace, but v3 site keys use the classic (non-enterprise)
+ * namespace and endpoint.
+ *
+ * @async
+ * @param {object} globals - An object containing read-only form instance, read-only target field instance and methods for form modifications.
+ * @returns {string} - The captcha token.
+ */
+function fetchCaptchaToken(globals) {
```
rismehta: We should put this af-core, so that even headless/EDS works

navneet-ag: @kartikey19427 why do we need this, enterprise did not require any custom functions for implementation?

navneet-ag:
Please add cypress test cases for runtime and authoring both
https://github.com/adobe/aem-core-forms-components/blob/master/ui.tests/test-module/specs/recaptcha/recaptcha.authoring.cy.js

ui.frontend/src/customFunctions.js
kartikey19427: Updated — removed the `submitForm` override. `RecaptchaImpl.getCaptchaDisplayMode()` now returns `invisible` for v3, so af-core's built-in `submitForm` auto-fetches the token natively (works on both the af-core and af-core-xfa runtimes). Also checked the old captcha implementation — the built-in already handles invisible/enterprise-score on the raw form, which is why the override's proxy workaround is no longer needed.

ui.frontend/src/customFunctions.js
kartikey19427: Updated — deleted these; they only existed to support the `submitForm` override, which has been removed.

kartikey19427:
@navneet-ag Updated — added authoring and runtime Cypress specs for v3: `recaptcha.authoring.cy.js` now covers selecting the v3 config (size field disables, like enterprise-score), and `recaptcha.runtime.cy.js` adds a v3 invisible-badge render test plus a submission test (skipped pending a real v3 site/secret key, since v3 has no universal public test key). Also added the v3 cloud-config + sample-page IT fixtures.

ui.frontend/src/customFunctions.js
kartikey19427: Good call. The canonical `fetchCaptchaToken` actually lives in `@aemforms/af-custom-functions` (af-core just looks it up by name), so I've opened a PR there to add v3 natively — this covers headless/EDS too: https://git.corp.adobe.com/livecycle/af-custom-functions/pull/15. Keeping the local wrapper in this PR until that ships and we bump the dependency here, so v3 doesn't regress in the meantime.

kartikey19427:
@navneet-ag The v3 Cypress tests are written and committed (5ed509afb): authoring (selecting the v3 config disables the size field) and runtime (v3 invisible-badge render + a `submission should pass for reCAPTCHA v3` test). Verified the v3 submission end-to-end locally — form submit → `POST /adobe/forms/af/submit` → 200.

PR #1986, opened 2026-09-15
feat(recaptcha): support reCAPTCHA v3 in adaptive forms (+ XFA no-iframe embed fixes)

kartikey19427:
## Summary

Adds reCAPTCHA **v3** support to the adaptive forms recaptcha component, targeting the `dev` branch. Same change set as #1977 (which targets `master`) — cherry-picked onto `dev`, net diff is identical.

- `RecaptchaImpl.getCaptchaDisplayMode()` returns `invisible` for v3, so af-core's built-in `submitForm` auto-fetches the token natively (both af-core and af-core-xfa runtimes) — no `submitForm` override.
- `fetchCaptchaToken` is a plain re-export of `cf.fetchCaptchaToken`; v3 support lands upstream in `@aemforms/af-custom-functions` (covers headless/EDS). Requires that release + a dependency bump before it is CI-green.
- Two XFA no-iframe embed fixes: register custom functions into the XFA runtime at bundle load; guard container submit/save handlers against a missing `state.events`.

## Testing

- Java unit tests for `getCaptchaDisplayMode` + updated exporter fixture.
- Frontend unit tests pass.
- Cypress: authoring (v3 config selection disables size) + runtime (v3 invisible-badge render and a `submission should pass for reCAPTCHA v3` test).
- Verified end-to-end locally: v3 form submit → `POST /adobe/forms/af/submit/... → 200`.


ui.frontend/src/index.js
```diff
@@ -37,6 +37,14 @@ import {customFunctions} from "./customFunctions";
  */
 window.guideBridge = new GuideBridge();
 
+// Register the repo's custom functions at bundle load so BOTH runtimes get them. In the XFA
+// bundle (main-xfa.js) this populates af-core-xfa's FunctionRuntime before the XFA form model is
+// created; the container view only ever registers into the non-XFA runtime via window.FormView, so
+// the XFA rule engine would otherwise see only its built-in defaults and have no fetchCaptchaToken
+// (breaking auto-fetch captcha - invisible/enterprise-score/v3 - on no-iframe XFA embeds).
+// registerFunctions is idempotent, so the later setupFormContainer registration is unaffected.
+FunctionRuntime.registerFunctions(customFunctions);
```
rismehta: This is not required, this already happens today, why did you add this ?

ui.tests/test-module/specs/recaptcha/recaptcha.runtime.cy.js
```diff
@@ -238,4 +254,37 @@ describe("Form Runtime with Recaptcha Input", () => {
...
+        cy.wrap().then(() => {
+            const [id] = Object.entries(formContainer._fields).find(([id]) => id.includes("captcha"));
+            // v3 has no visible challenge; the widget must render badge-only (invisible).
+            cy.get(`#${id} .cmp-adaptiveform-recaptcha__widget > div.g-recaptcha`)
+                .should('exist')
+                .and('have.class', 'g-recaptcha-invisible');
+        });
+    })
+
+    it("submission should pass for reCAPTCHA v3", () => {
```
rismehta, line 284: I dont see the usage of CYPRESS_RECAPTCHA_V3_API_KEY secret in the test ?

rismehta (changes requested):
check comments

ui.frontend/src/index.js
kartikey19427: On a no-iframe XFA form the rule engine runs on af-core-xfa's FunctionRuntime, which ships no default fetchCaptchaToken, so auto-fetch captcha (invisible v2 / enterprise-score / v3) failed at submit with fetchCaptchaToken is not defined.  
I tested the no iframe case for recaptcha v3 and enterprise without this change. Both were failing with no submit action. This line was actually needed.

ui.tests/test-module/specs/recaptcha/recaptcha.runtime.cy.js
kartikey19427: Earlier i was testing with google public keys(testing keys) for recaptcha V3. Since that always results in a constant score it will eliminate the need for secret key. But I have reverted that back now since it wasn't working. Should be present now.

ui.frontend/src/index.js
rismehta: This already happens here, it is redundant check, https://github.com/adobe/aem-core-forms-components/blob/82a3dad5d67630c7717b72d9825287b74cde4086/ui.frontend/src/utils.js#L312

This should not happen at multiple places, check the existing occurence and add it correctly

rismehta (changes requested):
check comments

ui.frontend/src/index.js
kartikey19427: You're right that registration shouldn't be scattered — but these two calls target different FunctionRuntime instances, so the index.js one isn't redundant. The -xfa build aliases @aemforms/af-core → @aemforms/af-core-xfa (webpack.common.cjs), so main.js and main-xfa.js each carry their own runtime. utils.js:312 runs inside setupFormContainer, invoked only via the global window.FormView (formcontainerview.js) — the non-XFA bundle — so it only ever populates af-core's runtime. The XFA rule engine runs on af-core-xfa, which never received customFunctions, so no-iframe XFA captcha submit failed with fetchCaptchaToken is not defined (verified: removing the index.js line reproduces it).

The clean consolidation would be to make index.js (the shared per-bundle entry) the single registration point and drop utils.js:312 — but that line predates this PR and sits in the universal form-bootstrap path (FormView.Utils.setupFormContainer is public API), so removing it inside a reCAPTCHA fix is a wider, cross-cutting change than I'd want to bundle here without full regression testing. I'd prefer to keep this PR minimal (index.js only) and do the consolidation as a follow-up — happy to raise a ticket. 

I feel claude is right here and removing any of them effects either the captcha or have cross repo implications(public API).

ui.tests/test-module/specs/recaptcha/recaptcha.runtime.cy.js
```diff
@@ -238,4 +252,51 @@ describe("Form Runtime with Recaptcha Input", () => {
...
```
navneet-ag, line 284: can we add a test case for reacaptcha v3failure too ?

kartikey19427: Done- added a v3 failure test case with threshold score 1. Same as enterprise.

navneet-ag: The newly added test case for failure is failing on the PR, please check it

ui.frontend/src/index.js
rismehta: Got the issue, this is specific to XFA. Can you move the registration here and remove it from utils.js

navneet-ag:
https://github.com/adobe/aem-core-forms-components/tree/master/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/recaptcha/v1/recaptcha This readme needs to be updated to tell about v3 support

ui.tests/test-module/specs/recaptcha/recaptcha.runtime.cy.js
```diff
@@ -238,4 +252,74 @@ describe("Form Runtime with Recaptcha Input", () => {
             });
         }
     })
+
+    it("should render reCAPTCHA v3 as an invisible badge", () => {
```
navneet-ag, line 268: Test cases for reCAPTCHA v3 should only work when FT is enabled. I cannot see any config update to enable this FT to run the test cases.

How are the test case for success even passing without enabling the Feature Toggle

rismehta (approved):
LGTM

ui.tests/test-module/specs/recaptcha/recaptcha.runtime.cy.js
kartikey19427: because FT is enabled on cq guides and that is merged and released

navneet-ag: That's not how feature toggles work. For this repo please enable FT on `it/config/src/main/content/jcr_root/apps/system/config/com.adobe.granite.toggle.impl.dev.DynamicToggleProviderImpl.cfg.json`
And the test case should have check so that it runs only when FT is on the latest addon is used for aem

navneet-ag: Reference PR https://github.com/adobe/aem-core-forms-components/commit/45b8dd029d2a65c77e78a0a49efee760d7984145#diff-9bc6007d162ace794f183d4b651328794b4ace20a44fe6a6ed351c6279340007

navneet-ag: Looks like you already aded the FT in json, just add a check in test case that FT is on and latest addon

kartikey19427: Yea I just enabled it. Waiting for the tests now.

kartikey19427: added the latest addon condition too

kartikey19427:
> https://github.com/adobe/aem-core-forms-components/tree/master/ui.af.apps/src/main/content/jcr_root/apps/core/fd/components/form/recaptcha/v1/recaptcha This readme needs to be updated to tell about v3 support

updated

